PR #18, opened 2021-03-15
GlobalVariable offset indirect indexing

jackoalan:
LLVM uses the G_GLOBAL_VALUE phony to access an element from a global aggregate when a register containing the base address is known. Indirect indexing address mode is ideal for implementing this.

```c
struct {
  char foo;
  char bar;
} myStruct;

void _start() {
  myStruct.foo = 0x42;
  myStruct.bar = 0x42;
}
```

Without this change, the offset is always zero so bar is aliased to foo.


mysterymath:
Thanks for the contribution!

If you don't mind my asking, how'd you hear about our project? I never expected to receive instruction selection PRs out of the blue!

jackoalan:
Coincidence mostly! I found johnwbyrd/llvm-mos last week and thought it would be useful to simultaneously learn llvm's backend and Super Famicom homebrew. That led me here. 

I'd like to get up to speed on the various ISel methods. The MOS architecture is rather minimal, so there aren't too many machine details to get distracted with.

PR #80, opened 2021-09-09
Add initial 65C02 support with X/Y stack instructions

jackoalan:
Supports four new instructions when built with -mcpu=mos65c02
* PHX
* PHY
* PLX
* PLY

Since the entire GPR register class is now directly usable with the stack, all stack utilization including callee saved registers, copy to V, and scavenger register saves conditionally use them.

Also cleans up target machine and subtarget implementations for CPU selection and the necessary clang driver changes to use `-mcpu=`


mysterymath:
Awesome PR! This should help quite a bit on the Commander X16 and friends, especially since A is usually live in to the prologue when CSRs are to be stored, and Y is usually free!

Looks good to me!

PR #81, opened 2021-09-11
Add 65C02 BRA instruction for near-range unconditional jumps

jackoalan:
The BRA instruction is an unconditional branch that uses PC-relative addressing. When -mcpu=mos65c02 is active, BRA is used as the default unconditional branch instruction.

The BranchRelaxation pass detects jumps that are too long for BRA and requests the target to insert a JMP as fallback.


mysterymath (approved):
LGTM!

mysterymath:
This PR was reverted in 34b64c2df58d04df91bb031206619087650d3f1f.

This broke the following integration test:
        SingleSource/Regression/C/gcc-c-torture/execute/GCC-C-execute-20071029-1
    
To debug this, run the test suite in verify mode (add
    -Wl,-mllvm,-verify-machineinstrs to the -DTEST_SUITE_ARCH_FLAGS cmake
    argument).
    
The error is as follows:
        LLVM ERROR: Attempting to emit BRA_Relative instruction but the Feature_Has65C02 predicate(s) are not met

Please create a PR that reverts the revert and fixes the code or test, depending on what the problem actually is.

mysterymath:
Giving the PR a quick once-over, I'd wager that this is due to the BRA logical instruction not being predicated on Feature_Has65C02. This might cause it to be selected via Pattern, even though once it was eventually expanded out to BRA_Relative, it'd be illegal.

jackoalan:
Thanks for the heads up! Good time for me to learn more capabilities of the generated pattern table. Looks you're right about the predicate though. TableGen adds a `OPC_CheckPatternPredicate` which appears to skip that match.

jackoalan:
> To debug this, run the llvm-test-suite in verify mode (add
> -Wl,-mllvm,-verify-machineinstrs to the -DTEST_SUITE_ARCH_FLAGS cmake
> argument).
Which optimization level was used to trigger this failure? It didn't happen for me on `-Os` but there are a ton of `!BundleNonDebug.isDebugInstr()` asserts in other tests.

mysterymath:
Here's the test run logs: https://github.com/llvm-mos/llvm-test-suite/actions/runs/1223860841
It appears to happen at all optimization levels.

The test steps all use the binaries from the latest release ZIP, while the verify steps build the compiler from scratch. Accordingly, 
there may be some stale binaries on your end. Particularly, make sure that lld is freshly built; since LTO is used by default, that's what'll actually generate most of the code.

PR #87, opened 2021-09-12
Refactor handling of EF_MOS_ flags

### LINKED ISSUE #31: Support feature bits in object files
Support has been laid out already for feature bits, so that the various 6502 variance can be marked with their own special instructions. However, there's no provision for making sure to link only compatible versions of the 6502.  It should be possible to query one of the flags in the ELF header of any object file, to see which processor features are supported.

jackoalan:
TableGen's SubtargetEmitter does not support bitwise feature values and also lacks visibility to other tools like lld and llvm-readobj. This refactor provides shared functions for handling EF_MOS_ logic across LLVM, including conflict detection and generation of flag strings.

Flag conflicts are detected per-IR-Function in the LLVM backend and per-object in LLD. This way, information about the first entity to cause a conflict is displayed to the user.

Crucially, ELF objects now have populated e_flags fields according to the llvm-mos ELF specification.

Closes #31

### DIFF
--- a/lld/ELF/Arch/MOS.cpp
+++ b/lld/ELF/Arch/MOS.cpp
@@ -11,6 +11,7 @@
 #include "Target.h"
 #include "lld/Common/ErrorHandler.h"
 #include "llvm/BinaryFormat/ELF.h"
+#include "llvm/BinaryFormat/MOSFlags.h"
 #include "llvm/Object/ELF.h"
 #include "llvm/Support/Endian.h"
 
@@ -26,6 +27,7 @@ namespace {
 class MOS final : public TargetInfo {
 public:
   MOS();
+  uint32_t calcEFlags() const override;
   RelExpr getRelExpr(RelType type, const Symbol &s,
                      const uint8_t *loc) const override;
   void relocate(uint8_t *loc, const Relocation &rel,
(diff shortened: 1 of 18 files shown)


mysterymath:
I should probably let @johnwbyrd take a look at this one; not sure I know the back half well to give it due diligence.

johnwbyrd:
Oh excellent, I am glad you took a crack at populating the e_flags fields.  I'll need a little time to do this review, as you've touched a lot of files.  Thanks for the effort; I'll get back to you.

jackoalan:
Cool! I'd also like some input on the logic for `checkEFlagsCompatibility` cause it's probably too lenient. An exhaustive truth table is probably overkill, but perhaps better detection of invalid BCD mixtures is in order.

johnwbyrd:
> Cool! I'd also like some input on the logic for `checkEFlagsCompatibility` cause it's probably too lenient. An exhaustive truth table is probably overkill, but perhaps better detection of invalid BCD mixtures is in order.

Note the ASCII graphic at https://docs.mamedev.org/techspecs/m6502.html .

jackoalan:
The BCD feature also presents an interesting question: Should it be disabled by default? Compiling an object with the default flags would have much more guaranteed compatibility with the newer subtargets that way.

`-mbcd` option would be the sensible way to enable it.

jackoalan:
> Note the ASCII graphic at https://docs.mamedev.org/techspecs/m6502.html .

This graphic doesn't really indicate where feature removals occur. No mention of the WDC variants either, particularly the 65816 in emulation mode.

johnwbyrd:
The logic in llvm/lib/BinaryFormat/MOSFlags.cpp needs a little more love, but I'm going to let this roll into main, because it's an improvement on the general state of things I think.

PR #90, opened 2021-09-19
Replace EM value

### LINKED ISSUE #86: ELF machine number mismatch
Official GNU binutils already recognizes 254 as `EM_MCS6502` as of late 2020: [elf/common.h](https://github.com/bminor/binutils-gdb/blob/master/include/elf/common.h), and the number was actually registered as of early 2019: [twitter](https://twitter.com/babbageboole/status/1094008708919582721). This project should probably be updated to use the existing assignment.

rdrpenguin04:
Fixes #86.

Essentially a find/replace of EM_MOS with EM_MCS6502 and an edit of the number (6502 -> 254; 0x1966 -> 0xFE in tests). Regression tests are still in progress locally.

### DIFF
--- a/lld/ELF/InputFiles.cpp
+++ b/lld/ELF/InputFiles.cpp
@@ -1632,7 +1632,7 @@ static uint16_t getBitcodeMachineKind(StringRef path, const Triple &t) {
   case Triple::mips64el:
     return EM_MIPS;
   case Triple::mos:
-    return EM_MOS;
+    return EM_MCS6502;
   case Triple::msp430:
     return EM_MSP430;
   case Triple::ppc:
--- a/lld/ELF/Target.cpp
+++ b/lld/ELF/Target.cpp
@@ -87,7 +87,7 @@ TargetInfo *elf::getTarget() {
     return getSPARCV9TargetInfo();
   case EM_X86_64:
     return getX86_64TargetInfo();
-  case EM_MOS:
+  case EM_MCS6502:
     return getMOSTargetInfo();
   }
   llvm_unreachable("unknown target machine");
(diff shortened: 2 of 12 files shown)


rdrpenguin04:
Tests just passed locally:
```
Testing Time: 5587.68s
  Unsupported      : 20918
  Passed           : 54169
  Expectedly Failed:    78
  ```

johnwbyrd:
"official."  Show me an alternative ELF specification for the 6502, please.

rdrpenguin04:
I'm having issues digging up the actual registration request (it was done via Xinuos after they stopped publishing updates to the specification; source [Twitter](https://twitter.com/babbageboole/status/1094089817070002176)), but I did find the [patch to binutils](http://patches-tcwg.linaro.org/patch/45921/) which I'd assume to be sourced from somewhere. Connor got lucky; he made his e_machine number request right as the transition was happening (source: [generic-abi mailing list](https://groups.google.com/g/generic-abi/c/qaPzp2lRzDA)), so the record is still easily accessible.

jackoalan:
Are we aware of any existing tools that use this magic number? If we adopt it, we run the risk of spec collisions. This is particularly concerning with the e_flags definitions since they now play a crucial role in sanity checking the linker.

rdrpenguin04:
This is the only project I know about that targets 6502 ELF.

johnwbyrd:
My current understanding is that we're going to wait until the gcc 65816 project is able to compile working code across some base of tests, after which point we will review the ELF integration issue in more detail.  I recommend that you join our Slack and engage in our real time discussions of our current calling convention and flags settings... more discussion would generally be productive.  Thanks for submitting this patch -- we may come back to it later.

rdrpenguin04:
To clarify, this issue has nothing to do with 65816, it is fully focused on 6502 ELF.

Could you send a link to your Slack server?

PR #105, opened 2021-10-11
Correct poor spilling and splitting scenario 2

beholdnec:
The previous merge could fail in regalloc due to generating "Yc = LDstk" patterns. I have patched it to avoid generating these patterns, albeit in a crude way. Please do not merge yet.


beholdnec:
Ok, ready to merge!

PR #119, opened 2021-11-21
LDZ pseudo instruction for zero-stores to imaginaries as STZ.

jackoalan:
For 65C02 subtarget, zero constants are selected as LDZ. After register
allocation, this is expanded as LDImm 0 for GPRs or STZImag8 for Imag8s.
The application of STZ_ZeroPage is effectively applied to zero-loads on
imaginary registers.


mysterymath (approved):
Awesome, LGTM!

PR #123, opened 2021-11-25
Lower G_FREEZE to COPY in legalizer.

jackoalan:
Since the InstructionSelector operates in reverse block order without
repetition, predecessor instructions that will soon be selected as COPY
will not be recognized in llvm::getDefSrcRegIgnoringCopies by default.

This gives getDefSrcRegIgnoringCopies a way to recognize a set of
alternative copy opcodes so it may function as expected.


jackoalan:
For reference, `cmp-select.ll` compiles as the following without this change:

```s
func:                                   ; @func
; %bb.0:                                ; %entry
	stx	mos8(__rc2)
	ldx	#1
	sec
	sbc	mos8(__rc2)
	beq	.LBB0_2
; %bb.1:                                ; %entry
	ldy	#0
	dey
	iny
	beq	.LBB0_3
	jmp	.LBB0_4
.LBB0_2:                                ; %entry
	ldy	#1
	dey
	iny
	bne	.LBB0_4
.LBB0_3:                                ; %select.false
	ldx	#2
.LBB0_4:                                ; %select.end
	txa
	rts
```

and with this change:

```s
func:                                   ; @func
; %bb.0:                                ; %entry
	tay
	stx	mos8(__rc2)
	lda	#1
	cpy	mos8(__rc2)
	beq	.LBB0_2
; %bb.1:                                ; %select.false
	lda	#2
.LBB0_2:                                ; %select.end
	rts
```

mysterymath (commented):
Can we achieve the same affect by lowering these to copies in the legalizer instead of the instruction selector?

If that works, it'd save a target hook, and it'd continue the recent vein of using the legalizer to normalize instructions. All of these instructions are exactly identical in behavior for our target, so making them formally equivalent (i.e. COPY) as early as possible simplifies all the later passes that would otherwise have to deal with them.

jackoalan:
> Can we achieve the same affect by lowering these to copies in the legalizer instead of the instruction selector?

That's a much better idea. Would also help target independent passes identify copies before instruction select.

jackoalan:
Lowering to COPY in the legalizer is not as trivial as I was expecting (assuming we want to keep the legalizer action logic).

Of course we could simply restrict the input types and connect those to a custom handler that replaces the opcode. Something along the lines of:

```cpp
getActionDefinitionsBuilder(G_FREEZE)
    .customFor({S1, S8, P})
    .unsupported();
```

Of course this undermines the type generality of G_FREEZE. At least in the G_SELECT condition operand case, it's always S1.

I'm at a loss for best alternatives though. All I can think of aside from a modified `getDefSrcRegIgnoringCopies` is a pre-select lowering pass of some kind.

mysterymath:
What's wrong with replacing the legalFor with customFor?

```
  getActionDefinitionsBuilder({G_IMPLICIT_DEF, G_FREEZE})
      .customFor({S1, S8, P})
      .widenScalarToNextMultipleOf(0, 8)
      .maxScalar(0, S8)
      .unsupported();
```

jackoalan:
> What's wrong with replacing the legalFor with customFor in the existing script?

Now that you mention it, nothing's wrong with that. I was concerned that G_FREEZE would be left intact in the widen/narrowing actions. However, it would be dominated by extension/truncation so `getDefSrcRegIgnoringCopies` would be stymied anyway.

I will make this change with test cases soon.

mysterymath:
>I was concerned that G_FREEZE would be left intact in the widen/narrowing actions.

There's no way for this to happen without triggering assertions in the legalizer. If the widening or narrowing materially alters the G_FREEZE, say by changing the type of one of its arguments, this is caught by the Observe, which considers it a new instruction. This means that the whole legalizer script will be run on the new G_FREEZE, starting from the top. If the new G_FREEZE were to make it all the way to the bottom of the script (say, by being a vector), it'd hit the unsupported at the bottom and crash the compiler.

jackoalan:
Thank you for clarifying the nature of the legalizer. I've confirmed all type cases with appropriate legalizer tests.

mysterymath (approved):
LGTM!

PR #125, opened 2021-11-29
Find return instruction if CSR restore point is set.

jackoalan:
To keep MachineCopyPropogation from removing copies in a restore block
that is not the same as the return instruction, traverse successors
until all reachable returns are found.

### DIFF
--- a/llvm/lib/Target/MOS/MOSFrameLowering.cpp
+++ b/llvm/lib/Target/MOS/MOSFrameLowering.cpp
@@ -116,6 +116,25 @@ bool MOSFrameLowering::spillCalleeSavedRegisters(
   return true;
 }
 
+template <typename F>
+static void visitReturnBlocks(MachineBasicBlock *MBB, const F &Func,
+                              DenseSet<MachineBasicBlock *> &VisitedBBs) {
+  if (!VisitedBBs.insert(MBB).second)
+    return;
+  if (MBB->isReturnBlock())
+    Func(*MBB);
+
+  // Follow branches in BB and look for returns
+  for (MachineBasicBlock *Succ : MBB->successors())
+    visitReturnBlocks(Succ, Func, VisitedBBs);
+}
+
+template <typename F>
+static void visitReturnBlocks(MachineBasicBlock *MBB, const F &Func) {
+  DenseSet<MachineBasicBlock *> VisitedBBs;
+  visitReturnBlocks(MBB, Func, VisitedBBs);
+}
+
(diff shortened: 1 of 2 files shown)


llvm/lib/Target/MOS/MOSFrameLowering.cpp
```diff
@@ -116,6 +116,28 @@ bool MOSFrameLowering::spillCalleeSavedRegisters(
   return true;
 }
 
+template <typename F>
+static void findReturnBlocks(MachineBasicBlock *MBB, const F &Func,
```
mysterymath, line 120: visitReturnBlocks.

llvm/lib/Target/MOS/MOSFrameLowering.cpp
```diff
@@ -116,6 +116,28 @@ bool MOSFrameLowering::spillCalleeSavedRegisters(
...
+                             DenseSet<MachineBasicBlock *> &VisitedBBs) {
+  if (!VisitedBBs.insert(MBB).second)
+    return;
+  if (MBB->isReturnBlock())
+    Func(*MBB);
+
+  // Follow branches in BB and look for returns
+  for (MachineBasicBlock *Succ : MBB->successors()) {
+    if (Succ == MBB)
```
mysterymath, line 129: Optional: This if is unnecessary, strictly speaking, as the VisitedBBs check should trigger. This may be slightly faster, but I doubt that really matters for our compiler in general.

llvm/lib/Target/MOS/MOSFrameLowering.cpp
```diff
@@ -152,14 +174,17 @@ bool MOSFrameLowering::restoreCalleeSavedRegisters(
...
     for (const CalleeSavedInfo &CI : CSI) {
-      MBB.rbegin()->addOperand(
-          *MBB.getParent(), MachineOperand::CreateReg(
-                                CI.getReg(), /*isDef=*/false, /*isImp=*/true));
+      MBB.rbegin()->addOperand(MachineOperand::CreateReg(
+          CI.getReg(), /*isDef=*/false, /*isImp=*/true));
     }
-  }
+  };
+  if (MBB.succ_empty())
```
mysterymath, line 184: No need to distinguish; we can just always call the function. The function will reduce to the former if there are no successors. (Although it'll do slightly more work in that case, O(1) performance differences are unlikely to matter much for llvm-mos.)

mysterymath (approved):
LGTM, sans a few nitpicks.

jackoalan:
Made requested changes

PR #130, opened 2021-12-23
Add initial debug info support with static stack.

jackoalan:
Formally enables debug info support. Debug locations and values are
already mostly functional. Also adds DWARF handling for frame objects
that are allocated on a function's static stack.


mysterymath (approved):
LGTM, thanks for the PR.

PR #133, opened 2021-12-31
Have CMake create mos-clang symlinks.

jackoalan:
This facilitates unambiguous access to LLVM-MOS via PATH.
As a bonus, clang recognizes the argv0 prefix and sets the default target to
mos.

Also disables alternate language bindings since they install artifacts
outside the normal prefix and risk conflicting with vanilla LLVM.


mysterymath (approved):
LGTM

PR #135, opened 2022-01-08
Perform legalizer info verification, fix `G_BRJT` rule.

jackoalan:
Call `LegacyLegalizerInfo::computeTables` and `LegalizerInfo::verify` as
part of the legalizer setup like other targets do. The only effect of
computing legacy tables is setting a flag indicating it has been called.
For opcodes without legalizer rules, this results in a more informative
assert:

```
LLVM ERROR: unable to legalize instruction: %178:_(s1) = G_FCMP floatpred(ule), %177:_(s64), %105:_ (in function: _ftoa)
```

as opposed to "backend forgot to call computeTables".

`G_BRJT` did not cover the scalar generic type with its rule, so a
compound predicate is used instead to ensure compatibility with the
emitted `G_TRUNC`.

### DIFF
--- a/llvm/lib/Target/MOS/MOSCombiner.cpp
+++ b/llvm/lib/Target/MOS/MOSCombiner.cpp
@@ -18,6 +18,7 @@
 #include "MCTargetDesc/MOSMCTargetDesc.h"
 #include "MOS.h"
 #include "MOSLegalizerInfo.h"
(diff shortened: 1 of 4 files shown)


mysterymath:
That's really weird; we used to get those error messages, and I could have sworn we used to call computeTables.

mysterymath:
LGTM though.

PR #136, opened 2022-01-10
legalize llvm.trap inrinsic

mrk-its:


llvm/test/CodeGen/MOS/trap.ll
```diff
@@ -0,0 +1,16 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py
+; RUN: llc -verify-machineinstrs -O0 --filetype=asm < %s | FileCheck %s
```
mysterymath, line 2: It's generally preferable to test changes like this as close to the pass as possible; otherwise, the test has a tendency to pick up irrelevant details and assert on them. Please do write this as a legalizer test; it can go in the existing legalizer.mir.

mysterymath (changes requested):

llvm/test/CodeGen/MOS/trap.ll
mysterymath: It should just be `G_INTRINSIC_W_SIDE_EFFECTS intrinsic(@llvm.trap)`.

mysterymath (approved):

PR #139, opened 2022-01-16
Handle global values in named zero-page sections.

jackoalan:
Applies the `mos8()` expression to any global value that is in a named
zero-page section (e.g. `__attribute__((section(".zp")))`). This works
even for external symbols that must be resolved by the linker, making
the user responsible for ensuring the linker will place the symbol
completely within the zero page.

MOSMCInstLower now always lowers to ZeroPage[X] where available.
Instructions that contain immediate address operands have a forced
relaxation performed just before emitting to the streamer. Fixup
relaxations occur as normal, including not relaxing for `mos8()` as
now placed for `section("<zero-page-section-name>")`.


jackoalan:
Confirmed passing llvm-test-suite

mysterymath (approved):
Nice, I was just thinking of doing something like this. Nice to see that emitting ZeroPage instructions in the compiler works with the assembler relaxation; that was a big question mark.

jackoalan:
In this case, the only "non standard" addition to assembler relaxation was early handling for immediate addresses. Normal assembler relaxation effectively occurs only for relocations.

Fortunately, it's fairly easy to invoke this after all MC lowerings including pseudo expansions.

PR #140, opened 2022-01-17
`.mos_addr_asciz` target directive with `R_MOS_ADDR_ASCIZ`

jackoalan:
To ensure consistency in the BASIC header on C64, a special relocation
type and assembly directive are provided to output the value as a
decimal ASCII string.

Syntax: `.mos_addr_asciz <expression>, <num-digit-chars>`

A fixed count of null terminators is output to reserve space for an
address that would be practical for the target platform. Prior to
relocation, the string "0" is emitted to the object file.


jackoalan:
Also renames `MOSELFStreamer` -> `MOSTargetELFStreamer` as it's easily confused with `MOSMCELFStreamer`

mysterymath (approved):
Ha, this is awesome; I was just planning to add a cheesy jump from a fixed start address.

LGTM; do add an entry somewhere in the llvm-mos.org wiki about this.

jackoalan:
Wiki articles:
https://llvm-mos.org/wiki/Assembler
https://llvm-mos.org/wiki/ELF_specification#Relocation_types

PR #147, opened 2022-01-30
Add MC instructions, relocations, and disassembler support for 65816.

jackoalan:
## New addressing modes

* Immediate16
* Relative16
* AbsoluteLong
* AbsoluteXLong
* Indirect (existing Indirect renamed to Indirect16)
* IndexedIndirect16
* IndirectLong
* IndirectIndexedLong

## New instructions

* Long addressing modes for ADC, SBC, AND, EOR, ORA, LDA, STA
* Long jump and subroutine instructions
* Immediate16 versions of applicable instructions
* SEP/REP

## Assembler support

* mos16() modifier to emit an 8-bit value as a 16-bit immediate operand.
* Zero-bank relaxation step to transform Addr16 -> Addr24 (may be
  chained with zero-page relaxation).

## Relocations

* R_MOS_PCREL_16
* Implement R_MOS_ADDR24 in lld
* Fix lld's ability to relocate PCREL (even though the assembler always does
  it in practice).

## Disassembler support

* Add Mapping symbols $ml, $mh, $xl, $xh to correctly decode 16-bit immediate
  operands. These set a simulated M and X flag high or low. Low states enable
  the Immediate16 decode tables.


mysterymath (commented):
Broadly LGTM for style and a cursory look at the content, but I don't actually know the 65816 very well.

jackoalan:
> Broadly LGTM for style and a cursory look at the content, but I don't actually know the 65816 very well.

Here's the primary resource used:
https://www.westerndesigncenter.com/wdc/documentation/w65c816s.pdf

I essentially extrapolated the concept of the zero-page to zero-bank, covering addresses from `$0-$ffff`. The 65816 lets you redefine which bank and zero page (actually "direct page") those addressing modes access.

I think dynamic pages and banks will be rather difficult to model in LLVM. Instead, instructions which do not have 24-bit long addressing modes available (shifts and increments) should trigger LLVM to copy to accumulator or imaginary for operation.

An analysis of existing logical instructions and how they should map to 65816 is in this spreadsheet:
https://docs.google.com/spreadsheets/d/1RS0qnQO1bN9BgDgk6sN65QzJoKghXvoEyPS-L7u4m-Q/edit?usp=sharing

mysterymath:
I think we're going to need to do a lot of experimentation with 65816 codegen; the available options are pretty dramatically different than the 6502.

For example, the WDC C compiler maintains the direct page pointer as the stack pointer, allowing one-byte accesses to stack-relative addresses. It's reasonably likely that this could be used to build a SPARC-style register windowing system, which would be a tremendously efficient way to pass values around the system. But this is one of a number of approaches that we'd want to explore to have a similar quality of code generation for the 65816 as we do for the 6502.

I'm also a bit concerned about harming 6502 code generation by adding a lot of 65816 conditionals to its implementation. We'd ideally need to keep the implementations reasonably separate, at least until we've got a really good idea of what good codegen for the 65816 looks like. At that point, once the implementation is more stable, it'd start making sense to build abstractions that could be shared between the two implementations.

Not too worried about things where we can already predict the shape of them; pretty much everything to do with linking and assembling is pretty set in stone as to how it would work, at least in broad strokes. But 65816 codegen is still a pretty green field as far as compilers go; not as green as the 6502, but greener than the 68k.

jackoalan:
Good points. I've marked this as a draft until we have a clearer idea if the codegen should take place in its own target implementation. There is certainly plenty of functionality that can be shared, but the maintenance issues surrounding a mess of conditionals would not be worth it.

mysterymath:
LGTM, I'll pass this off to @johnwbyrd

mysterymath (approved):

mysterymath:
It's been a while, so I'll go ahead and merge this.

johnwbyrd:
Thanks for merging this, @mysterymath .  Sorry I haven't had the time that this changeset deserved.

PR #156, opened 2022-02-15
Refactor

mjanos5:
Seems we have some infrastructure issue with windows tests, not related to my changes.


llvm/lib/Target/MOS/MCTargetDesc/MOSInstPrinter.cpp
```diff
@@ -31,27 +31,48 @@
...
+      CorrectOperands += *Operand;
     }
-    CorrectOperands += Letter;
   }
-  OS << CorrectOperands;
+
+  return CorrectOperands;
+}
+
+void MOSInstPrinter::printInst(const MCInst *MI, uint64_t Address,
```
mysterymath, line 72: This makes it far less clear what this routine is doing. (Although it already wasn't very clear; at least the mess was all in one place!)

It looks like the logic here is:
1) Print the instruction as-is to a string.
2) Replace the first two spaces with tabs.
3) Print the resulting string.

Accordingly, I'd expect this routine to be:
```
std::string Str;
raw_string_ostream StrStream(Str);
auto MnemonicInfo = getMnemonic(MI);
assert(MnemonicInfo.second && "Missing opcode for instruction.");
printInstruction(MI, Address, StrStream);
OS << convertSpacesToTabs(StrStream.str());
```

mysterymath (changes requested):

llvm/lib/Target/MOS/MCTargetDesc/MOSInstPrinter.cpp
mysterymath: On second thought, `convertSpacesToTabs` isn't a good name for this, since it doesn't just convert spaces to tabs. Instead, the intent is to normalize the whitespace, so maybe `normalizeWhitespace`.

mjanos5: I renamed some things as you asked. For the rest feel free to extend my solution, as I am only refactoring the existing code and can not execute tests so I want to avoid making functional changes. You also understand better what happens here. I think separating string creation and printing it to the stream could be a good idea, as the creation could be unit tested without having a real stream. Regarding the single big block, I find it more readable if the related things are separated by using empty lines.

mysterymath: As-is, I can't accept the PR, as it harms readability without improving functionality.

mjanos5: It is subjective what improves these things. Splitting the code to smaller, clean parts is an improvement for me and reduces the mess. Not to mention the formatting part where you had 3 instances of very similar code what I joined according to the DRY principle. You can continue my work as well if you feel value in it but prefer to implement something differently, and integrate it when it is fine for you. You agreed it was a mess even before it. If you are open to have contributors, you could also have some flexibility, especially in subjective things if the functionality is OK. As I mentioned I don't have the environment to make functional changes.

mysterymath: Yes, style issues fall somewhere on a spectrum of subjective to objective. But I'd tend to disallow any PR that doesn't provide some kind of plausibly objective benefit, as every change to a codebase incurs risk.

PR #173, opened 2022-03-25
Show absolute address in branch instruction disassembly

### LINKED ISSUE #172: Show absolute address for branch instructions
Currently we show relative address which is unreadable.

Let's do this conditionally in `MCInstPrinter`, taking into account existing `PrintBranchImmAsAddress` flag (and let's set this flag in lldb)

mrk-its:
fixes #172


mysterymath:
It looks like this is causing test failures in mos-relocs.s; would you take a look?

mrk-its:
> It looks like this is causing test failures in mos-relocs.s; would you take a look?
yes, taking a look.

mysterymath (approved):
Nice, this makes llvm-objdump -d much easier to read. LGTM.

lld/test/ELF/mos-relocs.s
```diff
@@ -50,8 +50,8 @@ relnext:
 # RELOCS-NEXT: 00000001 R_MOS_PCREL_8            .R_MOS_PCREL_8+0x4
 # RELOCS-NEXT: 00000003 R_MOS_PCREL_8            .R_MOS_PCREL_8
 # CHECK-LABEL: section .R_MOS_PCREL_8:
-# CHECK: bpl $2
-# CHECK: bpl $fc
+# CHECK: bpl $100c5
```
mrk-its, line 53: Why these sections are located above 0xffff here, @mysterymath?

mysterymath: The default ELF linker script built into the linker chooses 0x10000 as its start address.

mysterymath: It looks like we can actually change this if need be.

PR #211, opened 2022-08-09
Add smb and cmb opcodes for 65c02

mlund:
This adds the set/clear memory bit opcodes (smb, cmb) for 65c02. They do however conflict with automatically generated (and invalid) opcodes from 6502. I will likely need input on how to best solve this.

Tasks:

- [x] Add set memory bit opcodes (smb)
- [x] Add clear memory bit opcodes (cmb)
- [x] Assembler test
- [ ] Remove conflict with invalid 6502 opcodes


mlund:
@mysterymath I wonder if I should try to convert this into an issue? Currently invalid 6502 opcodes are being generated which occupy the slots where smb/cmb goes in newer processors. I think it would be useful to complete 65c02 support as it is the only processor still in production and (hopefully) used in the Commander X16.

mysterymath:
Yeah, please do file a separate issue. I'm not 100 percent on the details, so it'd be a good idea to get it in writing and track the work.

xfix:
I believe those instructions aren't a part of 65C02, but rather were a popular extension first introduced by R65C00/21 CPU, later integrated into W65C02S (which is used by Commander X16).

asiekierka:
You're correct; however, this still creates a decoding conflict, as the WDC 65C816 is based on the WDC 65C02, but (unlike the WDC 65C02) omits the Rockwell extensions.

mysterymath:
This was superceded by #296; closing.

Issue #190, opened 2022-06-26
Elide saving callee-saved registers if function doesn't return


atn34:
Consider this c program
```
char g(char, char);
  
void foo() {
        char x0 = 0;
        char x1 = 0;
        char x2 = 0;
        for (;;) {
                x0 = g(x1, x2);
                x1 = g(x0, x1);
                x2 = g(x1, x0);
        }
}
```
for which llvm-mos generates (with -O3)
```
        .text
        .file   "test.c"
        .globl  foo                             ; -- Begin function foo
        .type   foo,@function
foo:                                    ; @foo
; %bb.0:
        lda     mos8(__rc20)
        pha
        lda     mos8(__rc21)
        pha
        ldx     #0
        stx     mos8(__rc21)
        ldx     #0
.LBB0_1:                                ; =>This Inner Loop Header: Depth=1
        lda     mos8(__rc21)
        jsr     g
        sta     mos8(__rc20)
        ldx     mos8(__rc21)
        jsr     g
        ldx     mos8(__rc20)
        sta     mos8(__rc21)
        jsr     g
        tax
        jmp     .LBB0_1
.Lfunc_end0:
        .size   foo, .Lfunc_end0-foo
                                        ; -- End function
        .ident  "clang version 15.0.0 (https://www.github.com/llvm-mos/llvm-mos.git d8b5ba2066d1ee73305fbedab1a209163202b355)"
        .section        ".note.GNU-stack","",@progbits
        .addrsig
```

It seems to me that since this function doesn't return and we never restore the callee-saved registers __rc20 and __rc21, that we don't need to save them in the first place.

For some additional context I'm playing with [dodo](http://www.dodolabs.io/) where your whole game basically must fit in 0x1f00 bytes, so minimizing code size is my primary interest. Also I'm hoping to get around to adding dodo as a supported platform for llvm-mos-sdk if that's welcome!

mysterymath:
Sounds reasonable, since we don't need stack unwinders to work on the 6502. Shouldn't be too difficult to fix.

mysterymath:
It looks like there may be support already in LLVM for this: see determineCalleeSaves in TargetFrameLoweringImpl.cpp. It may just be a matter of turning it on.

gandhi56:
`determineCalleeSaves()` is already being called in the mos-zero-page-alloc pass. The allocations seem to be related to multiple calls to `g`.

gandhi56:
IR (1):
```
[...]

PR #227, opened 2022-10-01
[MOS][CodeGen] Enable skipping callee-saved registers

gandhi56:
Avoid saving callee-saved registers for functions
attributed with NoReturn and NoUnwind. This patch
allows for more imaginary registers to be available for use by the callee.


mysterymath:
LGTM! Thanks for working on this!
Closes #190.

PR #254, opened 2022-11-11
Fix "void main" violating the C and C++ standards.

pfusik:
Common error. :) `clang` issues a warning. `clang++` issues an error.

BTW. Maybe the resulting disassembly should be updated as well? For instance, aren't `jsr`+`rts` coalesced into `jmp` now? Don't know how to recreate: `-O2 -S -fno-lto` emits `jmp puts` and `mos-c64-clang -O2 -Wall hi.c` unrolls.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ LLVM-MOS is a LLVM fork supporting the MOS 65xx series of microprocessors.
 ```C
 #include <stdio.h>
 
-void main(void) {
+int main(void) {
   printf("HELLO, 6502!\n");
 }
 ```


mysterymath:
It's generated with -Os, as that generally produces the best code for 6502 platforms. -O2 and -O3 trade too much space away for time.

pfusik:
Thanks!
The output did not change. Now I see the `jsr` comes from `asm`, so can't coalesce with `rts`.

pfusik:
By "did not change" I meant same code for `void main`. With `int main` there's `ldx #0` `txa` before the `rts`.

mysterymath:
Ah, yeah, sorry for the premature squash. I'll clean up the Readme when I get a sec.

larsbrinkhoff:
It's somewhat jarring that the return type is now int, but there's no return statement.

pfusik:
@larsbrinkhoff That complies to the C and C++ standards. `void main` was not.

larsbrinkhoff:
You are correct.  I argue it's jarring (a subjective value statement), not against the standard.

PR #255, opened 2022-11-11
Update output for "int main".

pfusik:


pfusik:
This even looks better for a 6502 programmer, because no missed opportunity for `jsr`+`rts` -> `jmp`.

pfusik:
The C standard says `main` returns zero if there's no `return` statement.

Issue #216, opened 2022-08-09
Implement `wai` and `stp` instructions


atn34:
These are instructions specific to the wdc 65c02: https://en.wikipedia.org/wiki/WDC_65C02#Low-power_modes, but that can't be used in assembly files currently.

mysterymath:
Note that this task requires adding parsing and machine code generation for these instructions to LLVM-MOS; it's not just a matter of turning them on.

mysterymath:
See also #54.

alexschrod:
While we're waiting for these instructions to be added proper, is there any way to just *force* an opcode into the generated assembly? All I need is to stuff a `$DB` byte into where this `STP` should be:
```
.globl debug
debug:
    STP
    RTS
```
I tried writing `#$DB` in place of the `STP` but it didn't end up in the final executable.

The way I'm working around this now is by putting `NOP` in place of `STP` and then manually hex editing the final executable to put the right opcode in, but this is obviously very tedious and time consuming.

atn34:
You can use the `.byte` directive, .e.g `.byte $CB` for `wai`

Issue #212, opened 2022-08-09
Missing address modes for certain 65c02 opcodes


mlund:
The 65c02 and derived CPUs have instructions with address modes that currently cannot be added. The Branch on Bit (Re)Set instructions `bbr0 zp, offset` where the `offset` is a relative 8-bit address is one example. I tried something like this:
``` cpp
def ZeroPageRelative : AddressingMode {let OperandsStr = "$param1 , $param2";
   let InOperandList = (ins addr8:$param1, pcrel8:$param2);
}
```
but the assembler currently cannot handle more than one argument, see [discussion on Slack](https://llvm-mos.slack.com/archives/C03NLRDB45D/p1657665050637599). To fully support 65c02 (`cx16` target) and derived processors (`mega65` target), this will need to be resolved.

Issue #54, opened 2021-06-08
Fill out instruction sets for 65c02, 65ce02, 65816, and other compatibles


johnwbyrd:
Although I've sketched in an outline of processor features at https://github.com/llvm-mos/llvm-mos/blob/main/llvm/lib/Target/MOS/MOSDevices.td , it needs to be fleshed out with actual new instructions in MOSInstr*.td, and test cases added to the assembler for each of these variants.

A good overview of 65xx variants is at https://docs.mamedev.org/techspecs/m6502.html .

mlund:
Some additional processor information that may be useful:

- https://gitlab.com/jespergravgaard/kickassembler65ce02.
- https://airtable.com/shrRRDDvm3dvzV2Y1

(The 45GS02 would be particularly interesting for the sdk's `mega65` target.)

mlund:
Issue seems related to #94.

matthiaskramm:
For 6502x, good sources are
http://www.oxyron.de/html/opcodes02.html
https://csdb.dk/getinternalfile.php/222714/NoMoreSecrets-NMOS6510UnintendedOpcodes-20212412.pdf 
https://www.pagetable.com/?p=39

PR #296, opened 2023-06-08
Add HuC6280 CPU target; 6502X, R65C02, W65C02, and W65C816 opcodes.

asiekierka:
This pull request adds (assembler/target-level) support for the HuC6280 CPU target, which is a purpose-specific variant of the 65C02 designed by Hudson Soft for use in the PC Engine; given that it's clocked at 7.16 MHz, it may prove itself to be a surprisingly interesting target for llvm-mos... In addition, it fleshes out support for the following missing/incomplete classes of opcodes:

* 6502X opcodes (so-called "illegal" opcodes),
* Rockwell and WDC 65C02 opcodes,
* WDC 65816 opcodes.

Affected issues:

* Supersedes https://github.com/llvm-mos/llvm-mos/pull/211 (RMB/SMB opcode PR),
* Resolves https://github.com/llvm-mos/llvm-mos/issues/216 ,
* Resolves https://github.com/llvm-mos/llvm-mos/issues/212 - though maybe not satisfactorily,
* Mostly resolves https://github.com/llvm-mos/llvm-mos/issues/54 - the 65EL02 and SWEET16 remain unimplemented, but the 65EL02 is a virtual machine specific to a Minecraft mod which has been abandoned for one decade now, while the SWEET16 would require special handling.

There are some things to be resolved still:

- The ELF specification needs to be amended to include EF_MOS_ARCH_HUC6280;
  - Some other CPUs could also be added to the specification: the CSG 4510 (with two custom opcodes of its own), C64DTV (with a legalized subset of "illegal" opcodes, plus its own custom opcodes), and the Renesas 740 (an expanded 6502 - though it has different opcode sets depending on the specific chip...).
- ~~The conflict from https://github.com/llvm-mos/llvm-mos/pull/211 needs to be fully resolved~~ This has now been done; if I did my research correctly, it appears that WDC dropped Rockwell's extensions for the 65816 specifically;
- Some assemblers (like 64tass) prefer to represent BBR/BBS/RMB/SMB opcodes as `RMB 0, $ea` rather than `RMB0 $ea`. Should llvm-mos go down that path as well?
- The backend doesn't emit these new instructions yet, but this can be resolved separately. For HuC6280 in particular, it was essential for bringup to get some of them (memory control, CPU speed control) working at the assembler level.
  - There is some low-hanging fruit, but I'd like to discuss them further on the Discord and/or such.


llvm/lib/Target/MOS/AsmParser/MOSAsmParser.cpp
```diff
@@ -648,33 +649,37 @@ class MOSAsmParser : public MCTargetAsmParser {
     */
     // First, the mnemonic goes on the stack.
     Operands.push_back(MOSOperand::createToken(Mnemonic, NameLoc));
-    bool FirstTime = true;
```
asiekierka, line 651: Modifications here were necessary to support multiple-argument opcodes (Rockwell's "branch bit reset/set" opcodes, as well as Hudson Soft's block move and bitwise test opcodes).

llvm/lib/Target/MOS/MOSInstrFormats.td
```diff
@@ -57,7 +72,7 @@ class Inst<string asmstr> : Instruction {
   let mayStore = true;
   let hasSideEffects = true;
 
-  field bits<32> SoftFail = 0;
+  field bits<56> SoftFail = 0;
```
asiekierka, line 75: Hudson's block move instructions are 7 bytes long.

asiekierka:
Actually adding them to the IR backend can be done in a separate step.

mysterymath:
Would you mind marking this one draft until you think it's ready to go in as-is? I can review it along the line, but with the pace of commits, it still looks very WIP, so it'll be otherwise unclear when I should merge it.

mysterymath:
Ah thanks, I'll start reviewing this then.

asiekierka:
One more note. Before merging, please do remember that changes would be required to match in the ELF specification:

* Adding EF_MOS_ARCH_HUC6280; this may also be a good opportunity to add EF_MOS_ARCH values for other 6502 derivatives not previously covered (some that come to mind: CSG 4510, C64DTV, 65SC02 which is a WDC 65C02 minus Rockwell extensions),
* Removing EM_MOS_R65C02 from the bits implied by EM_MOS_W65816 - the WDC 65816 does not provide the Rockwell 65C02 extensions.

johnwbyrd:
I'm not personally familiar with the 6502 variants in question, but if you think the ELF spec should be expanded to cover them, then I'm certainly willing to let you take a crack at modifying it.  If you do make the modifications, we'd have to bump the version number to 0.2.1, since the EM_MOS_R65C02 change is a breaking change.

asiekierka:
I've just been doing research in the area, I'm not an expert with them myself.

In general, I would apply the following specific changes, working from the assumption that ELF machine flags are meant to correspond to CPU features specifically:

* Add `EF_MOS_ARCH_HUC6280`, `0x00000800`, `HuC6280 only instructions and addressing modes, including SAX, SAY, SXY, ST. This bit may be set to one only if the EM_MOS_6502, EM_MOS_65C02, and EM_MOS_R65C02 bits are set to one.` This corresponds with [this HuC6280 opcode matrix](https://www.chrismcovell.com/PCEdev/HuC6280_opcodes.html).
* Amend `EM_MOS_W65816` to remove `EM_MOS_R65C02` from the required bits to set - the WDC 65816 does not implement Rockwell's 65C02 extensions, having its own instruction additions under the same opcodes.
* Add `EF_MOS_ARCH_65DTV02`, `0x00001000`, `65DTV02 only instructions - BRA, SAC and SIR - as defined in [64tass reference manual](http://tass64.sourceforge.net/#opcodes-65dtv02). This bit may be set to one only if the EM_MOS_6502 bit is set to one.`
* Add `EF_MOS_ARCH_4510`, `0x00002000`, `CSG 4510 only instructions - MAP - as defined in [64tass reference manual](http://tass64.sourceforge.net/#opcodes-4510) This bit may be set to one only if the EM_MOS_6502, EM_MOS_65C02, and EM_MOS_R65C02 and EM_MOS_65CE02 bits are set to one.`
  * The 65CE02 has the MAP opcode slot (`0x5C`) [reserved for future expansion](https://www.pagetable.com/c64ref/6502/?cpu=65ce02&tab=2#AUG), according to documentation - so whether or not the 4510 should count as a separate architecture, or if the 65CE02 architecture should simply provide MAP in its place, is up for discussion if we want to conserve bits.
* Amend `EM_MOS_W65C02` to remove `EM_MOS_R65C02` from the required bits to set - there are CPUs which have WDC, but not Rockwell, 65C02 extensions (commonly, the 65SC02 is said to be such, but I cannot find a consistent source on which 65C02 variants have which Rockwell/WDC extensions).
* Amend `EM_MOS_65EL02` to remove `EM_MOS_R65C02` from the required bits to set - it's an enhanced 65C02, yes, but it does not feature the Rockwell extensions.
  * Cosmetic: Link to `http://www.eloraam.com/nonwp/redcpu.php` for the 65EL02 - this is the authoritative reference as made by the virtual machine's original developer.
  * Cosmetic: Link to `http://bigfootinformatika.hu/65el02/archive/65el02_instructions.txt` for the 65EL02 as an instruction set additions reference.
  * Note: The 65EL02 is not, actually, fully 6502-compatible - it has a bug with its implementation of the stack pointer (it's off by +1 relative to a real 6502), so a lot of `TSX`/`TXS`-reliant code will not work on it out of the box.

The following platforms were found in the course of my research which I am not sure what to do about:

* The SPC700, which is Sony's heavily 6502-inspired CPU - while a lot of things crucial from the perspective of an LLVM backend are the same, the entire opcode matrix is laid out differently, and there are some differences ([here](https://wiki.superfamicom.org/uploads/spc700_opcode_matrix_1b.png) is an SPC700 opcode matrix using 6502-style mnemonics). I'm not confident on proposing an answer to this, but if I would, I would either define a new `EF_MOS_ARCH_SPC700` that doesn't require any of the other bits, or a new machine type entirely (which would then be out of scope).
* The Mitsubishi/Renesas 740, which are 65C02 derivatives. They have a lot of variants according to Wikipedia, though, and I haven't looked deeper into it yet.

asiekierka:
A more careful observation showed that stack-relative addressing modes for the 65816 were not added previously - but, as they introduce a conflict with the 65CE02, I will keep them to a separate PR.

asiekierka:
I have finished updating the ELF specification on the wiki.

johnwbyrd:
ELF spec LGTM.

PR #297, opened 2023-06-12
Support emitting CLA/CLX/CLY on HuC6280.

asiekierka:
This is the first HuC6280-specific optimization I am proposing for llvm-mos, and it's by far the simplest one to implement.

The HuC6280 is an extended Rockwell-standard 65C02 which provides [some additional opcodes](http://shu.emuunlim.com/download/pcedocs/pce_cpu.html). Among them is `CLA`, `CLX`, and `CLY`: they act like `LDA #$00`, `LDX #$00` and `LDY #$00` respectively, taking the same amount of clock cycles, but with two differences:

1. They take up only one byte instead of two (but the same number of clock cycles, that is 2 clock cycles),
2. They do not modify the NZ flags.

As llvm-mos already includes a pseudo-instruction to "load zero into a register", it was possible to extend it to recognize this scenario.

### DIFF
--- a/llvm/lib/Target/MOS/MOSInstrInfo.cpp
+++ b/llvm/lib/Target/MOS/MOSInstrInfo.cpp
@@ -1031,12 +1031,17 @@ void MOSInstrInfo::expandLDImm16Remat(MachineIRBuilder &Builder) const {
 void MOSInstrInfo::expandLDZ(MachineIRBuilder &Builder) const {
   auto &MI = *Builder.getInsertPt();
   Register DestReg = MI.getOperand(0).getReg();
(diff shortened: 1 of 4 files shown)


mysterymath (approved):
nit: Change the description of LDZ in MOSInstrPseudos.td to just say "Load zero into GPR". With this change There's enough unique ways to lower this now that it's not a good idea to list them explicitly.

Also, this could use an end-to-end MIR test to demonstrate that a LDZ can be lowered to e.g. a CLA.

Otherwise, LGTM!

mysterymath (approved):
Yep, that'll do!

Issue #217, opened 2022-08-09
Add CPU type for the RP2A03 (NES)


mysterymath:
A CPU type should be added for the Ricoh 2A03 (and implicitly the PAL 2A07) used in the NES. The only difference is that the D flag has no effect on arithmetic.

Note that this shouldn't involve actually disabling the CLD and SED instructions; the D flag is still there, and those instructions do set and reset it. This is visible via PHP;PLA. The only difference is that the code generator need not concern itself with preparing the D flag, e.g, in interrupt handlers.

mysterymath:
Deprioritizing due to the deprioiritization of llvm-mos/llvm-mos-sdk#72.

johnwbyrd:
FWIW the ELF specification already contemplates the Ricoh 2A03... https://llvm-mos.org/wiki/ELF_specification#Machine_type_field

PR #298, opened 2023-06-12
Finish 65816 opcode support, add 65EL02 opcodes, 65DTV02, 4510 subtargets and opcodes.

asiekierka:
This PR completes the work started in https://github.com/llvm-mos/llvm-mos/pull/296 with the following changes:

* Finish correcting the outputted ELF machine type to match the ELF specification draft, version 0.2.1 (as discussed in the above PR),
* Add support for 65816 stack-relative opcodes, thus completing 65816 support at the machine code/assembler level (I didn't realize they were missing in #296 - oops!),
* Add support for 65EL02 opcodes,
* Add a 65DTV02 subtarget, complete with opcodes,
* Add a 4510 subtarget, complete with opcodes,
* Consistently rename the "mosw65el02" target to "mos65el02", matching other non-WDC subtargets.

There is are two reasons for which this is marked as a *draft* pull request:

1. ~~I did not manage to figure out how to cleanly define an InstAlias of the form `NOP` -> `EOM`, required for the CSG 4510's `MAP` opcode (which [defines a memory mapping configuration](http://www.devili.iki.fi/Computers/Commodore/C65/System_Specification/Chapter_2/page38.html), ended by `EOM`).~~ Done!
2. ~~The only remaining issue related to LLVM-side subtarget definitions appears to be https://github.com/llvm-mos/llvm-mos/issues/217 - and it would be fairly easy to just add a `mosrp2a03` subtarget that covers it.~~ Left for a separate PR.


mysterymath:
> 1. I did not manage to figure out how to cleanly define an InstAlias of the form `NOP` -> `EOM`, required for the CSG 4510's `MAP` opcode (which [defines a memory mapping configuration](http://www.devili.iki.fi/Computers/Commodore/C65/System_Specification/Chapter_2/page38.html), ended by `EOM`).
Can we just use something like a MnemonicAlias to map all seen `EOM`s to `NOP`? They should be functionally equivalent, right, it's just that a NOP is treated differently in the context of a MAP sequence?

> 2. The only remaining issue related to LLVM-side subtarget definitions appears to be [Add CPU type for the RP2A03 (NES) #217](https://github.com/llvm-mos/llvm-mos/issues/217) - and it would be fairly easy to just add a `mosrp2a03` subtarget that covers it.
Since there's already a separate issue for this, it seems ok to me to just leave this out of scope for this PR; no need to do everything all at once.

mysterymath (commented):
Broadly LGTM, pending a solution for NOP/EOM.

asiekierka:
Your proposed solution seems to work just fine, thank you!

mysterymath (approved):

PR #299, opened 2023-06-16
Support emitting TXY/TYX on W65816/65EL02.

asiekierka:
As I was exploring the area anyway - was wondering how to implement the HuC6280's SAX/SAY/SXY register *swap* opcodes into the backend, especially SXY sounds like it could be useful in some edge cases (advice/suggestions welcome!) - I decided to add TXY/TYX emitting support for W65816 and 65EL02 subtargets.


llvm/test/CodeGen/MOS/late-opt-65816.mir
```diff
@@ -0,0 +1,30 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py
+# RUN: llc -mtriple=mos -mcpu=mosw65816 -run-pass=mos-late-opt -verify-machineinstrs -o - %s | FileCheck %s
+---
```
mysterymath, line 3: Can you add a test for the CMPTermZ changes as well?

mysterymath (approved):
Nice!

mysterymath:
> (advice/suggestions welcome!)
If an when I'm actually able to finish the 6502-specific register allocator I'm working on, this would be very helpful in implementing the reg-to-reg shuffles as part of that. Otherwise, the only thing I can think of off the bat is to add a pass before COPY lowering to recognize swaps of those registers via a temporary that is killed during the swap;  these could safely be replaced with the corresponding S instructions. But I'd expect the savings there to be pretty marginal compared to the work involved.

llvm/test/CodeGen/MOS/late-opt-65816.mir
asiekierka: Done!

asiekierka:
The other usecase I saw for `SXY` in particular is handling `X->Y` and `Y->X` transfers when the source operand is killed (there's a convenient `KillSrc` argument to `copyPhysReg`) - would you say that such a scenario occurs at all in llvm-mos generation, though?

mysterymath:
> The other usecase I saw for `SXY` in particular is handling `X->Y` and `Y->X` transfers when the source operand is killed (there's a convenient `KillSrc` argument to `copyPhysReg`)

Ohhhh, there's no direct TXY on that chip! Then yes, definitely, that's the ideal lowering in copyPhysReg, and having that available would tend to mean it'd get a lot of use. (There generally aren't many copies of the same value floating around at the same time, since it's slightly wasteful, so kills of copy sources are very common.)

PR #300, opened 2023-06-17
Support more optimized X<->Y register transfer on 65C02 and HuC6280.

asiekierka:
As discussed in https://github.com/llvm-mos/llvm-mos/pull/299 , the PC Engine's HuC6280 lacks an X->Y/Y->X transfer opcode, but in many cases (source register killed) this can be worked around with an X<->Y swap.

In addition, the 65C02 can do `PHX` / `PLY` instead of `PHA` / `TXA` / `TAY` / `PLA`; this has been implemented as well.

To be resolved: 

- [x] Reflect added copy strategy in the cost calculation (`MOSRegisterInfo::getRegAllocationHints` -> `copyCost`).


llvm/lib/Target/MOS/MOSInstrLogical.td
```diff
@@ -411,6 +411,16 @@ let Predicates = [HasW65816Or65EL02] in {
 def TX : MOSTransfer<XY, XY>;
 }
 
+let Predicates = [HasHUC6280] in {
+// SAX, SAY, SXY
+def SWAP : MOSLogicalInstr {
+  dag OutOperandList = (outs GPR:$dst);
```
mysterymath, line 417: Since this reads and writes both registers, it needs to be modeled to use and define both registers. This can be done by having two uses and two defs, with the corresponding uses and defs tied together. Otherwise, it appears that this instruction only modifies one of its operands, which makes calls like `MI.modifiesRegister()` unsound.

asiekierka: I'm sorry! That was a little embarassing...

Is this correct? I'm not sure about my usage of the `kill` flag here - my assumption is that it is correct due to the existence of virtual registers; I'm more confident in my use of `undef` for defining the other register in the context of modeling a copy using a swap.

```
    ; CHECK-NEXT: $y, $x = SWAP killed $x, killed undef $y
    $y = COPY killed $x
```

mysterymath: Lol, don't be embarassed; this is the quickest I've ever seen anyone pick up LLVM, this shit is hard!
Yeah, that looks right to me. The value of $y doesn't matter for the operation, and both values have their live ranges end due to the immediately-following defs.

asiekierka:
I'm not confident enough to come up with a better way to calculate costs than mere averaging here. However, I did observe a `PHA` / `TXA` / `TAY` / `PLA` pattern being emitted, which is unnecessary on 65C02 and derivatives, as one can emit `PHX` / `PLY` instead.

asiekierka:
One other note - do we want to split `PH` and `PL` into NMOS (`A`, `P` only) and CMOS (`A`, `X`, `Y`, `P`) variants, as was done with `IN` and `DE`?

mysterymath:
> One other note - do we want to split `PH` and `PL` into NMOS (`A`, `P` only) and CMOS (`A`, `X`, `Y`, `P`) variants, as was done with `IN` and `DE`?

Yeah, please do.

llvm/lib/Target/MOS/MOSRegisterInfo.cpp
```diff
@@ -815,12 +815,26 @@ int MOSRegisterInfo::copyCost(Register DestReg, Register SrcReg,
...
     if (STI.hasW65816Or65EL02()) {
       // TXY, TYX
       return 3;
     }
-    // May need to pha/pla around; avg cost 4
-    return 4 + copyCost(DestReg, MOS::A, STI) + copyCost(MOS::A, SrcReg, STI);
+    int XYCopyCost;
+    if (STI.has65C02()) {
+      // PHX/PLY, PHY/PLX
+      XYCopyCost = 8;
```
mysterymath, line 827: nit: should be 9: 2 bytes + 7 cycles

mysterymath (approved):

mysterymath:
> I'm not confident enough to come up with a better way to calculate costs than mere averaging here.

Averaging is the best that can be done with the interface here; eventually this should become an input to a big SDCC/Philip Krause-style dynamic programming algorithm; in that case we'll know explicitly which registers are free when calculating the cost, which should allow producing exact values.

asiekierka:
Okay, I split PH and PL into PH_CMOS and PL_CMOS variants (AXYP) and the regular variants (AP only).

mysterymath (approved):

PR #302, opened 2023-06-17
Support emitting INC A/DEC A on 65C02.

asiekierka:
Continuing with the low-hanging fruit.


llvm/lib/Target/MOS/MOSInstrLogical.td
```diff
@@ -355,8 +355,8 @@ let mayLoad = true in {
 
 multiclass MOSIncDec {
   def NAME : MOSLogicalInstr {
-    dag OutOperandList = (outs XY:$dst);
-    dag InOperandList = (ins XY:$src);
+    dag OutOperandList = (outs GPR:$dst);
```
mysterymath, line 358: I did a little double-take here; the modelling here is tricky, since this is only the correct register class for this instruction on 65c02 and friends. That means if this instruction were ever involved in register allocation, it could cause INA to be selected on NMOS 6502, which doesn't correspond to a real instruction.

I think it's sufficient to document the invariant here that the argument must never be set to A on targets without an INA.

llvm/lib/Target/MOS/MOSInstrInfo.cpp
```diff
@@ -1054,6 +1054,7 @@ void MOSInstrInfo::expandLDZ(MachineIRBuilder &Builder) const {
 
 void MOSInstrInfo::expandIncDec(MachineIRBuilder &Builder) const {
```
mysterymath, line 1055: The INC/DEC pseudos also aren't quite ideal anymore for targets with INA, since they clobber C and V to accomodate `CLC;ADC#1` and `CLC;ADC#$ff`. We should split them into INC_NMOS and INC_CMOS, no action required on this change, but it's a good idea if you have time.

asiekierka: I'm going to mark this as draft until your solution is done; it is certainly a better approach.

llvm/lib/Target/MOS/MOSInstrLogical.td
asiekierka: Wouldn't it be slightly better to split `IN`/`DE` into a NMOS and CMOS variant, like you suggested with the INC/DEC pseudos?

mysterymath: Eh, yeah, let's. There's a key tradeoff in compilers: you want things to be as normalized as possible to reduce the number of cases you have to handle, but we probably shouldn't go as far as to remove safety guarantees.

asiekierka: Done, for both.

I appreciate the mentorship, by the way.

mysterymath (approved):

PR #303, opened 2023-06-18
Support BIT immediate on 65C02 to set V.

asiekierka:
Another low-hanging fruit; saves 1 byte and 3 cycles per call, plus doesn't require the `__set_v` symbol.


llvm/lib/Target/MOS/MOSRegisterInfo.cpp
```diff
@@ -842,6 +842,7 @@ int MOSRegisterInfo::copyCost(Register DestReg, Register SrcReg,
         getMatchingSuperReg(SrcReg, MOS::sublsb, &MOS::Anyi8RegClass);
     Register DestReg8 =
         getMatchingSuperReg(DestReg, MOS::sublsb, &MOS::Anyi8RegClass);
+    const int BitCost = STI.has65C02() ? 4 : 8;
```
mysterymath, line 845: This should save 2 cycles and 1 byte, right? The costs here are a straight sum of cycles and bytes, so I'd expect the costs to be 4 and 7.

Ideally we'd tune the costs more principally, taking into account preference for size vs speed, but a simple sum at least ensures that a point on the Pareto frontier is chosen.

mysterymath (approved):

llvm/lib/Target/MOS/MOSRegisterInfo.cpp
asiekierka: Correct, but this is actually due to an issue which I only noticed now and I'll need to resolve in another PR: the HuC6280's CPU has an additional one-cycle cost to what is apparently all non-immediate/implied addressing modes.

mysterymath (approved):

PR #304, opened 2023-06-18
Fix ELF relocation and relaxation for multi-operand 65C02/HuC6280 opcodes.

asiekierka:
* Introduce R_MOS_IMM16 ELF relocation - see below;
* Fix emitting operand relocations for multi-operand instructions;
* Fix relaxation of HuC6280's TST instruction - the code assumed only one-operand instructions need relaxation, which was no longer true after adding this instruction;
* Introduce tests for the two fixes above.

I'd like to raise attention to the added R_MOS_IMM16 ELF relocation, as it requires a change to the ELF specification. My intended usecase here is using the HuC6280 block copy instruction for copying data in crt0, as in:

```asm
    ; Copy __data_size bytes from __data_start to __data_load_start.
    tii __data_start, __data_load_start, #__data_size
```

This requires a symbol relocation in the form of a 16-bit immediate value.

### DIFF
--- a/llvm/lib/Target/MOS/MOSInstrFormats.td
+++ b/llvm/lib/Target/MOS/MOSInstrFormats.td
@@ -130,67 +130,80 @@ class Addr16Operand<string name> : MOSAsmOperand<name>;
 class Addr24Operand<string name> : MOSAsmOperand<name>;
 
 /// This operand will only match a value from 0 to 255 inclusive.
-def imm8 : Operand<i32> {
+class imm8at<int offset> : Operand<i32> {
   let ParserMatchClass = ImmediateAsmOperand<"Imm8">;
-  let EncoderMethod = "encodeImm<MOS::Imm8, 1>";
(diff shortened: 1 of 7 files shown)


llvm/lib/Target/MOS/MOSInstrFormats.td
```diff
@@ -362,6 +376,26 @@ def IndirectRStackRelativeY : AddressingMode {
   let InOperandList = (ins addr8:$param);
 }
 
+def ZeroPageAt2 : AddressingMode {
```
asiekierka, line 379: This is a little clunky, but I don't think there's a strong need to clean this up right now - as far as I know, the `TST` instruction is the only multi-operand instruction with a choice of multiple addressing modes like that.

mysterymath (approved):

asiekierka:
Should I add the `R_MOS_IMM16` change to the ELF specification, then? Plus, is this considered a breaking or non-breaking change (so I can increment the version accordingly)?

PR #305, opened 2023-06-18
Add PCE platform support to llvm-mlb.

asiekierka:
For https://github.com/llvm-mos/llvm-mos-sdk/pull/121.

### DIFF
--- a/llvm/tools/llvm-mlb/llvm-mlb.cpp
+++ b/llvm/tools/llvm-mlb/llvm-mlb.cpp
@@ -250,10 +250,9 @@ int main(int argc, char **argv) {
           uint8_t Bank = (Address >> 16) & 0xFF;
           int32_t DefaultOffset = -1; // Default offset.
           SmallString<32> Type;
-          fprintf(stderr, "%08lX\n", Address);
 
           if (Group == 0x00) {
-            // 0x00XXXXXX, bank 00-7F - card ROM bank addresses
+            // 0x00XXXXXX - base PC Engine bank addresses
             if (Bank >= 0x00 && Bank <= 0x7F) {
               Type = "PcePrgRom";
               DefaultOffset = Bank * 0x2000;
@@ -264,7 +263,7 @@ int main(int argc, char **argv) {
               return false;
             }
           } else if (Group == 0x01) {
-            // 0x01XXXXXX, bank 00-EF - card RAM bank addresses
+            // 0x01XXXXXX - card/CD-ROM RAM bank addresses
             if (Bank >= 0x80 && Bank <= 0x87) {
               Type = "PceCdromRam";
               DefaultOffset = (Bank - 0x80) * 0x2000;
@@ -278,13 +277,11 @@ int main(int argc, char **argv) {
               return false;
             }
           } else if (Group <= 0x11) {
-            fprintf(stderr, "a\n");
             // (0x02-0x11)XXXXXX, bank 40-7F - card ROM bank address (SF2 mapper)
             if (Bank < 0x40 || Bank >= 0x80)
               return false;
             Type = "PcePrgRom";
             DefaultOffset = (Group * 64 + (Bank - 0x40)) * 0x2000;
-            fprintf(stderr, "b %08X\n", DefaultOffset);
           } else {
             return false;
           }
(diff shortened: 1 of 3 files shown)


llvm/tools/llvm-mlb/llvm-mlb.cpp
```diff
@@ -63,92 +112,169 @@ int main(int argc, char **argv) {
...
+      for (const ELFSymbolRef Sym : O->symbols()) {
+        SymbolRef::Type Type = unwrapOrError(Sym.getType(), InputFilename);
+        uint64_t Address = unwrapOrError(Sym.getAddress(), InputFilename);
+        StringRef Name = unwrapOrError(Sym.getName(), InputFilename);
+        uint64_t Size = Sym.getSize();
+
+        if (Type == SymbolRef::ST_File)
+          continue;
+        if (Type == SymbolRef::ST_Unknown)
+          if (Name.startswith("__") && !Name.startswith("__rc"))
```
asiekierka, line 244: I've observed that this effectively prevents emitting symbols for some compiler-library symbols (like `__call_indir` or `__set_v`) - maybe the bound should be made stricter?

asiekierka:
I think it's unlikely I'm going to break this further; I'm happy with the LMA/VMA shape - unless you have any objections to it (I documented the layout in the parent PR in llvm-mos-sdk).

llvm/tools/llvm-mlb/llvm-mlb.cpp
mysterymath: Would you mind filing a bug for this? This is surprising; I had expected that the type check for ST_Unknown would preclude either of them from being included here, but it may be necessary to examine the section the symbols come from.

llvm/tools/llvm-mlb/llvm-mlb.cpp
```diff
@@ -25,6 +27,50 @@
...
+                 clEnumValN(MLBPlatform::NES, "NES", "NES"),
+                 clEnumValN(MLBPlatform::PCE, "PCE", "PCE")),
+      cl::NotHidden,
+      cl::cat(MLBCategory));
+
+  enum class MLBOutputFormat {
+    MLB
+  };
+
+  cl::opt<MLBOutputFormat> OutputFormat(
```
mysterymath, line 51: Can we strip this out until we need it? It's generally not a good idea to introduce flags before there's at least two alternatives; they increase the surface area of legal invocations of the tool, but don't increase it's functionality.

llvm/tools/llvm-mlb/llvm-mlb.cpp
```diff
@@ -25,6 +27,50 @@
...
+      cl::init(MLBOutputFormat::MLB),
+      cl::values(clEnumValN(MLBOutputFormat::MLB, "MLB", "Mesen Label File.")),
+      cl::NotHidden,
+      cl::cat(MLBCategory));
+
+  cl::list<std::string> ClInputFilenames(cl::Positional, cl::OneOrMore,
+                                    cl::desc("<input ELF files>"),
+                                    cl::cat(MLBCategory));
+
+  cl::opt<std::string> ClOutputFilename("o", cl::init(""),
```
mysterymath, line 62: Can we add a test that covers this option?

llvm/tools/llvm-mlb/llvm-mlb.cpp
```diff
@@ -63,92 +113,223 @@ int main(int argc, char **argv) {
 
     StringRef OutputBase = InputFilename;
     OutputBase.consume_back(".elf");
-    OutputBase.consume_back(".nes");
-    SmallString<32> OutputFilename = OutputBase;
-    OutputFilename += ".mlb";
+    bool IsInputExtNES = OutputBase.consume_back(".nes");
+    bool IsInputExtPCE = OutputBase.consume_back(".pce");
+    if (Platform == MLBPlatform::Auto) {
```
mysterymath, line 118: Can we cover these behaviors by tests? Should be able to re-use test asserts, just make sure that various invocations all Filecheck to the same thing.

mysterymath (approved):

asiekierka:
One more thing that occured to me regarding the LMA format.

The group (bits 31-24) "0x00" and "0x01" technically point to the same area in memory (bank 0x68 of the cartridge's address space can be RAM on some types of configurations and ROM on others); however, Mesen's labels require the distinction. Is there a better way to distinguish between them which doesn't require such an artificial split at the LMA level?

mysterymath:
Not that I know of; if you knew all the symbols would be in the same section, you could use the section name, but otherwise encoding it into the address is the best practice I know of.

asiekierka:
Okay, then.

mysterymath (approved):

PR #306, opened 2023-06-20
Add bounds support to lld custom output formats.

asiekierka:
This adds two new arguments: "start" and "length" to TRIM and FULL.

For example, "FULL(sec, 0, 4)" will only output 4 bytes, starting from position 0, of the section "sec". If "sec" is smaller than 4 bytes, it will only output as many bytes as the size of "sec".

As discussed on the llvm-mos Discord, with some thoughts after having used it.

### DIFF
--- a/lld/ELF/ScriptParser.cpp
+++ b/lld/ELF/ScriptParser.cpp
@@ -1285,15 +1285,23 @@ MemoryRegionCommand *ScriptParser::readMemoryRegionCommand(StringRef tok) {
   if (isFull == -1)
     return nullptr;
 
+  Expr regionStart = nullptr;
+  Expr regionLength = nullptr;
   expect("(");
   StringRef name = next();
+  if (consume(",")) {
+    regionStart = readExpr();
+    if (consume(",")) {
+      regionLength = readExpr();
+    }
+  }
   expect(")");
   auto iter = script->memoryRegions.find(name);
   if (iter == script->memoryRegions.end()) {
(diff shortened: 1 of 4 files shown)


lld/ELF/ScriptParser.cpp
```diff
@@ -1285,15 +1285,23 @@ MemoryRegionCommand *ScriptParser::readMemoryRegionCommand(StringRef tok) {
...
     return nullptr;
 
+  Expr regionStart = nullptr;
+  Expr regionEnd = nullptr;
   expect("(");
   StringRef name = next();
+  if (consume(",")) {
+    regionStart = readExpr();
+    if (consume(",")) {
+      regionEnd = readExpr();
```
asiekierka, line 1295: Question: Do we prefer "start, end" semantics, or (more akin to how other places in the linkscript operate) "start, length" semantics?

asiekierka: I've decided to switch to "start, length" semantics. They're a bit more intuitive, given that the values in use are unsigned.

mysterymath: SGTM.

mysterymath (approved):
Please do add documentation to the [wiki](https://llvm-mos.org/wiki/Linker_Script) for this. In particular, it's important to note that the start expression is relative to the beginning of the region, not an absolute LMA.

asiekierka:
Done.

PR #310, opened 2023-07-01
Initial support for building libcxx/libcxxabi on MOS.

asiekierka:
```
$ mkdir build-libcxx
$ cd build-libcxx
$ cmake -G Ninja -C ../clang/cmake/caches/MOSLibCXX.cmake -S ../runtimes
```

Proof of concept; I focused on getting it to compile only, though `hello-new.cc` seems to work (albeit with `heap_limit` etc. needing to be replaced by `__heap_limit` etc. - I assume libcxx's C stdlib wrappers break that approach).

Above all,  I'd like to hear your feedback.


mysterymath:
The first thing that immediately comes to mind is how to test this; libcxx comes with a pretty extensive test suite, and we should probably try to get as much of it working as we can, otherwise we're never going to have any idea whether this stuff actually functions properly.

Like the end-to-end test suite, this will probably involve hooking it up to our simulator target and using a test runner. We'll also have go to through and scrub out anything float-related; I presume you've already done so, since this compiles, but there may be tests for this stuff that casually relies on floats.

Along those lines, building might be a bit of a strong term here; especially if this was done with `mos-sim-clang`, since that defaults to output IR. Nothing will actually complain if the backend can't lower that to 6502 assembly; I suspect that may be the case for quite a lot of this.

asiekierka:
>Like the end-to-end test suite, this will probably involve hooking it up to our simulator target and using a test runner.

Probably. There's more concerns though - I expect a lot of the test binaries to not fit in the 64K address space at all. C++ in a complete STL sense is, in my opinion, not suitable for a target like this, but...

>We'll also have go to through and scrub out anything float-related

I've only scrubbed it out insofar as it prevented compilation, so not very much; my main focus with getting a larger C++ library going was to allow the compile-time powers of C++, which don't rely on the machine backend as much, and can make use of floats as well if one is careful.

>Nothing will actually complain if the backend can't lower that to 6502 assembly; I suspect that may be the case for quite a lot of this.

Probably.

libcxx/include/__functional/hash.h
```diff
@@ -50,6 +50,20 @@ __loadword(const void* __p)
 template <class _Size, size_t = sizeof(_Size)*__CHAR_BIT__>
 struct __murmur2_or_cityhash;
 
+template <class _Size>
+struct __murmur2_or_cityhash<_Size, 16>
+{
+    _LIBCPP_HIDE_FROM_ABI _LIBCPP_DISABLE_UBSAN_UNSIGNED_INTEGER_CHECK
+    _Size operator()(const void* __key, _Size __len) const {
+      // TODO: This is a stub implementation. Implement something better.
```
mysterymath, line 58: Let's call these TODO(llvm-mos); otherwise it will be more difficult to find them again. Consider filing issues for these too; that will make them easier to track.

libcxx/include/cstdio
```diff
@@ -157,7 +157,9 @@ using ::perror _LIBCPP_USING_IF_EXISTS;
 
 using ::fopen _LIBCPP_USING_IF_EXISTS;
 using ::freopen _LIBCPP_USING_IF_EXISTS;
+#ifndef _LIBCPP_HAS_NO_REMOVE
```
mysterymath, line 160: Is there something specific that makes remove different than all the other functions here? It seems odd to introduce a _LIBCPP_HAS_NO_REMOVE if not... maybe just #ifndef __mos__ otherwise?

libcxx/include/string
```diff
@@ -981,7 +981,7 @@ public:
     size_type __str_sz = __str.size();
     if (__pos > __str_sz)
       __throw_out_of_range();
-    __init(__str.data() + __pos, std::min(__n, __str_sz - __pos));
+    __init(__str.data() + __pos, std::min(__n, static_cast<size_t>(__str_sz - __pos)));
```
mysterymath, line 984: These seem strange; __str_sz, __pos, and __n should all be of type size_type, which I'd expect prima facie to be size_t. This shouldn't be promoted if arithmetic is performed on them, so I'd expect the result of the subtraction to already be size_t. Did you get an error message that describes what the type actually is in these examples?

libcxx/src/chrono.cpp
```diff
@@ -6,6 +6,9 @@
 //
 //===----------------------------------------------------------------------===//
 
+// FIXME: This file does not compile correctly under LLVM-MOS.
```
mysterymath, line 9: TODO(llvm-mos):

libcxx/src/include/to_chars_floating_point.h
```diff
@@ -60,11 +60,19 @@ struct _Floating_type_traits<float> {
 
     using _Uint_type = uint32_t;
 
+#if defined(__mos__)
+    static constexpr uint32_t _Exponent_mask             = (1ul << _Exponent_bits) - 1;
```
mysterymath, line 64: UINT32_C(1); then we probably shouldn't need the #if.

libcxx/include/string
asiekierka: Yes. It claims to be unable to build a `std::min` between `unsigned short` (`__n`) and `unsigned int` (`__str_sz - __pos`).

kassane:
Missing #422 reference.

Issue #207, opened 2022-07-26
Mem-family instructions have terrible lowering


mysterymath:
The `mem...` family instructions have pretty terrible lowering for small sizes; they always perform a libcall. Just setting up the arguments for the libcall can be more code than doing the operation outright.

The Legalizer pass actually already has a mechanism for dealing with this, it's just subtle, and the interface doesn't announce itself. We should configure it's optimization for the sizes of operations that should be handled inline on the 6502.

PR #311, opened 2023-07-01
Implement memory function lowering and HuC6280 block copy handling

asiekierka:
My first look at the Legalizer pass. Resolves https://github.com/llvm-mos/llvm-mos/issues/207

Also adds support for inlining `G_MEMCPY` and `G_MEMSET` using HuC6280's `TII` block copy instructions, complete with all the special handling this needs to not be too miserable an experience for the end user.

TODO:

- [x] Learn more about the Combiner pass and decide if HuC block copy emitting should continue to live in a separate pass of its own.
- [x] Write tests.
- [x] Give the code one more look.

TODO (after this PR):

- [x] Implement `G_MEMMOVE` HuC6280 lowering - only when are known at compile time and can either be proven not to overlap or the overlap direction can be discerned.

Notes:

- Memory instructions have sub-optimal/curious lowering for `volatile` copies - they are emitted as batches of `G_LOAD` followed by `G_STORE`, which prohibits interleaving them by the MachineScheduler. More interestingly, this means that a `memcpy(volatile, volatile, 4);` (lowered) will read and write values in a different order than `memcpy(volatile, volatile, 3);` (libcall).


mysterymath:
> Memory instructions still have bad lowering - they are emitted as batches of G_LOAD followed by G_STORE, which leads to the register allocator putting all of them in the zero page.

This is very surprising; after instructions are selected for these, the MachineScheduler should interleave them to reduce the register pressure. (We don't have a pipeline, so MachineScheduler is the dedicated "reorder instructions to decrease register pressure" pass.)  We should definitely look into why it isn't; there may be something slightly different we can emit to get it to do so, or we may be hitting a limitation of this pass that we can relax.

asiekierka:
Ah, I see why.

```c
#include <stdint.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

#define L 4
#define a ((volatile uint8_t*) 0x2200)
#define b ((volatile uint8_t*) 0x2300)

int main() {
  memcpy(b, a, L);
  return 0;
}
```

The key is the `volatile` - normally, this wouldn't matter for a `memcpy` call, but when `InstCombinerPass` replaces such a `memcpy` with an `s32` `G_LOAD`/`G_STORE`, it preserves the `volatile` attribute of the pointer - it is also preserved when the `G_LOAD`/`G_STORE` is legalized into multiple `s8` `G_LOAD`/`G_STORE`s.

So the issue happens at the `Legalizer` pass.

asiekierka:
I've updated the notes in the PR comment to clarify the situation. Sorry!

llvm/lib/Target/MOS/MOSHuCBlockCopy.cpp
```diff
@@ -0,0 +1,348 @@
...
+// Part of LLVM-MOS, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This file defines the MOS pass to lower G_MEMCPY/G_MEMSET/G_MEMSET_INLINE
+// opcodes to special HuC6280 variants prior to legalization, as information
+// which is difficult to extract mid-legalization (whether operands point to
+// opaque constants) is required.
```
mysterymath, line 12: This should be fairly straightforward to obtain during legalization, since uses are always legalized before defs. When a G_MEM<xxx> is seen by the legalizer, from the point of view of it and all of its arguments, that whole DAG is before legalization.

I haven't read this carefully yet, since it seems liable to change after my InstCombine change, but ideally this kind of logic would live in the legalizer. The way we've handled target-specific abstract operations is by introducing new target-specific G_<...> tags to lower them to; they can then usually get picked up fairly easily in the register allocator, and can do things like, e.g., encode directly that the arguments are constant. The various addressing mode opcodes (e.g., G_LOAD_IDX) are a good example of this; they encode an abstract operation that can be done on the target chip, potentially (but not necessarily) separate form the specific instruction that implements it (e.g., a G_STORE_IDX G_INC G_LOAD_IDX triple may be selected as a INC Abs,X).

asiekierka: Fixed - though this means G_LOAD/G_STORE pair detection needs to be moved to the combiner.

llvm/lib/Target/MOS/MOSCombiner.cpp
```diff
@@ -309,6 +327,115 @@ bool MOSCombinerHelperState::applyCMPZZero(MachineInstr &MI,
...
+  for (uint32_t I = 1; I < Bytes; I++) {
+    if (((Value >> (I * 8)) & 0xFF) != (Value & 0xFF)) {
+      return false;
+    }
+  }
+  return true;
+}
+
+  // G_LOAD/G_STORE pair => G_MEMCPY_INLINE
+bool MOSCombinerHelperState::matchLoadStoreToMemcpy(
```
mysterymath, line 341: This transformation isn't legal if the G_LOAD and G_STORE are both volatile, since there may be a C sequence point between the load and the store in the original program. If so, the C abstract machine requires that the volatile load be complete at that point, and that the volatile store has not yet begun. Converting this to memcpy would imply that there is not point in the resulting program where this is true.

llvm/lib/Target/MOS/MOSISelLowering.h
```diff
@@ -67,6 +67,16 @@ class MOSTargetLowering : public TargetLowering {
...
   bool isZExtFree(Type *SrcTy, Type *DstTy) const override;
 
+  EVT getOptimalMemOpType(const MemOp &Op,
+                          const AttributeList &FuncAttributes) const override {
+    return MVT::i8;
+  }
+
+  LLT getOptimalMemOpLLT(const MemOp &Op,
+                         const AttributeList &FuncAttributes) const override {
+    return LLT::scalar(8);
```
mysterymath, line 77: Nice! Love discovering that LegalizerHelper already has the logic we need buried inside.

llvm/lib/Target/MOS/MOSCombiner.cpp
asiekierka: Fixed.

asiekierka:
Okay, everything should be taken care of now! (Probably "almost everything", but.)

llvm/lib/Target/MOS/MOSInstrLogical.td
```diff
@@ -532,6 +532,21 @@ let Predicates = [Has65C02] in {
...
 
+//===---------------------------------------------------------------------===//
+// HuC Block Copy Instructions
+//===---------------------------------------------------------------------===//
+
+let Predicates = [HasHUC6280] in {
+  // (TII, TDD) abs, abs, imm16
+  def HuCMemcpy : MOSLogicalInstr {
+    dag InOperandList = (ins addr16:$source, addr16:$dest, imm16:$length,
+                         imm8:$descending);
```
mysterymath, line 543: i1imm

llvm/lib/Target/MOS/MOSMCInstLower.cpp
```diff
@@ -669,6 +681,7 @@ bool MOSMCInstLower::lowerOperand(const MachineOperand &MO, MCOperand &MCOp) {
 
   switch (MO.getType()) {
   default:
+  printf("Oper %d\n", MO.getType());
```
mysterymath, line 684: Debug print? Debug print.

llvm/test/CodeGen/MOS/combiner.mir
```diff
@@ -190,3 +263,50 @@ body: |
...
+body: |
+  bb.0.entry:
+    ; CHECK-LABEL: name: store_repeating_to_memset
+    ; CHECK: [[GV:%[0-9]+]]:_(p0) = G_GLOBAL_VALUE @a
+    ; CHECK-NEXT: [[C:%[0-9]+]]:_(s8) = G_CONSTANT i8 1
+    ; CHECK-NEXT: [[C1:%[0-9]+]]:_(s16) = G_CONSTANT i16 4
+    ; CHECK-NEXT: G_MEMSET [[GV]](p0), [[C]](s8), [[C1]](s16), 0 :: (store (s32) into @a)
+    ; CHECK-NEXT: RTS implicit [[GV]](p0)
+    %1:_(p0) = G_GLOBAL_VALUE @a
+    %0:_(s32) = G_CONSTANT i32 16843009
```
mysterymath, line 309: Add a comment with the hex here to make it clearer what's happening. (Or ideally, use hex literals if that works.)

mysterymath (commented):
Went through everything except the legalizer changes; I'll tackle that a bit later.

llvm/lib/Target/MOS/MOSCombiner.cpp
```diff
@@ -309,6 +330,121 @@ bool MOSCombinerHelperState::applyCMPZZero(MachineInstr &MI,
...
+  
+  auto StoreSrcReg = MI.getOperand(0).getReg();
+  auto Length = MRI.getType(StoreSrcReg).getSizeInBytes();
+  auto *SrcMemOperand = MIPrev->memoperands()[0];
+  auto *DstMemOperand = MI.memoperands()[0];
+
+  auto LengthReg = MRI.createGenericVirtualRegister(LLT::scalar(16));
+  B.buildConstant(LengthReg, Length);
+
+  B.buildInstr(MOS::G_MEMCPY_INLINE)
```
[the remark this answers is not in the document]
asiekierka: You'd think, but:

```llc: [...]/llvm-mos/llvm/lib/CodeGen/MachineInstr.cpp:228: void llvm::MachineInstr::addOperand(llvm::MachineFunction &, const llvm::MachineOperand &): Assertion `(MCID->isVariadic() || OpNo < MCID->getNumOperands() || Op.isValidExcessOperand()) && "Trying to add an operand to a machine instr that is already done!"' failed.```

File: llvm/lib/Target/MOS/MOSCombiner.cpp [hunk 8]
```diff
@@ -309,6 +330,121 @@ bool MOSCombinerHelperState::applyCMPZZero(MachineInstr &MI,
...
+
+  auto StoreSrcReg = MI.getOperand(0).getReg();
+  auto Length = MRI.getType(StoreSrcReg).getSizeInBytes();
+
+  auto ValueReg = MRI.createGenericVirtualRegister(LLT::scalar(8));
+  B.buildConstant(ValueReg, Value);
+  auto LengthReg = MRI.createGenericVirtualRegister(LLT::scalar(16));
+  B.buildConstant(LengthReg, Length);
+
+  B.buildInstr(MOS::G_MEMSET)
```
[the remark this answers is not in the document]
asiekierka: Not quite, but close.

llvm/lib/Target/MOS/MOSLegalizerInfo.cpp
```diff
@@ -1686,6 +1713,247 @@ bool MOSLegalizerInfo::selectIndirectIndexedAddressing(LegalizerHelper &Helper,
...
+    auto SrcValue = getUInt64FromConstantOper(Src.value());
+    if (!SrcValue.has_value())
+      return false;
+    if (MRI.getType(SrcReg).getSizeInBytes() != 1)
+      return false;
+  }
+  if (MI.getOpcode() == MOS::G_MEMMOVE) {
+    int OperandOrder = compareOperandLocations(Src.value(), Dst.value());
+    // TODO: Handle case when two G_MEMMOVE destinations cannot alias.
+    if (OperandOrder == -2)
```
mysterymath, line 1854: Prefer to have this return an std::optional<int>; the indended sematics of -2 is less clear than nullopt

llvm/lib/Target/MOS/MOSLegalizerInfo.cpp
```diff
@@ -1686,6 +1713,247 @@ bool MOSLegalizerInfo::selectIndirectIndexedAddressing(LegalizerHelper &Helper,
...
+
+  // Proceed with the custom lowering.
+  if (IsSet) {
+    // Emit a G_STORE, then set Src = Dst, Dst = Dst + 1, Len = Len - 1.
+    auto StoreReg = MRI.createGenericVirtualRegister(LLT::scalar(8));
+    Builder.buildConstant(StoreReg,
+                          getUInt64FromConstantOper(Src.value())
+                            .value() & 0xFF);
+    Builder.buildStore(StoreReg, DstReg,
+                       *MF.getMachineMemOperand(MachinePointerInfo(),
```
mysterymath, line 1924: This should use the destination's machine pointer info; it's the same pointer. If we can, it's a good idea to avoid loosing pointer provenance like this; it makes machine scheduling later more difficult (i.e. this store becomes a global barrier for code motion of loads and stores).

llvm/lib/Target/MOS/MOSLegalizerInfo.cpp
```diff
@@ -1686,6 +1713,247 @@ bool MOSLegalizerInfo::selectIndirectIndexedAddressing(LegalizerHelper &Helper,
...
+
+    Src = Dst;
+    Dst = offsetMachineOperand(Dst.value(), 1);
+    Len = offsetMachineOperand(Len.value(), -1);
+  }
+
+  // Note that Descending transfers must be done in backwards order.
+  if (KnownLen <= BytesPerTransfer) {
+    uint64_t AdjOfs = Descending ? (KnownLen - 1) : 0;
+    Builder.buildInstr(MOS::HuCMemcpy)
```
mysterymath, line 1936: Similarly, these instructions load and store memory, so they need their own memory operands, based on those from the original generic instruction. Otherwise, code motion passes will pessimize the aliasing possibilities of the instruction.

mysterymath (approved):

llvm/lib/Target/MOS/MOSLegalizerInfo.cpp
asiekierka: Done, I think?

PR #312, opened 2023-07-02
Model bytes and cycles separately in cost calculations.

asiekierka:
Mostly inspired by how easy it was to make a off-by-N mistake when writing the summed values by hand; but this also allows distinguishing between byte and cycle calculations for different optimization strategies. It also handles averaging better (by providing effective fixed-point representations of bytes and cycles - so "5.5 cycles" can be represented internally).

Setting as draft as I'm not 100% sure on the implementation details yet, but I think the general idea is sound. Not that I think it will impact optimization strategies much, at least not the current ones.


llvm/lib/Target/MOS/MOSCycleCost.h
```diff
@@ -0,0 +1,76 @@
+//===-- MOSCycleCost.h - MOS Cycle Cost structure ---------------*- C++ -*-===//
```
mysterymath, line 1: MOSInstrCost, definitely; it handles both cycles and bytes.

mysterymath (commented):
Yep, this is a very good idea.

llvm/lib/Target/MOS/MOSCycleCost.h
asiekierka: Fixed!

llvm/lib/Target/MOS/MOSInstrCost.h
```diff
@@ -0,0 +1,76 @@
...
+  PreferBytes,
+  PreferCycles,
+  Average
+};
+
+class MOSInstrCost {
+public:
+  MOSInstrCost() : Bytes(0), Cycles(0) {}
+
+  MOSInstrCost(int32_t Bytes, int32_t Cycles, int Multiplier = 256)
```
mysterymath, line 31: The multiplier-less version of this should be private; this would make it impossible to mix multipliers.

llvm/lib/Target/MOS/MOSInstrCost.h
```diff
@@ -0,0 +1,76 @@
...
+
+#ifndef LLVM_LIB_TARGET_MOS_MOSINSTRCOST_H
+#define LLVM_LIB_TARGET_MOS_MOSINSTRCOST_H
+
+#include "llvm/CodeGen/MachineFunction.h"
+#include <cstdint>
+
+namespace llvm {
+
+enum class MOSInstrCostMode {
```
mysterymath, line 21: Since the name is already a prefix, this should be a member of MOSInstrCost: MOSInstrCost::Mode.

llvm/lib/Target/MOS/MOSInstrCost.h
```diff
@@ -0,0 +1,76 @@
...
+    return MOSInstrCost(Left.Bytes / Right, Left.Cycles / Right, 1);
+  }
+
+  int64_t value(MOSInstrCostMode Mode = MOSInstrCostMode::Average) const;
+
+private:
+  int32_t Bytes, Cycles;
+};
+
+MOSInstrCostMode getMOSInstrCostModeFor(const MachineFunction &MF);
```
mysterymath, line 72: Since the name is already a prefix, this should be a static member of MOSInstrCost:
MOSInstrCost::getModeFor();

mysterymath (approved):

llvm/lib/Target/MOS/MOSInstrCost.h
asiekierka: Fixed.

llvm/lib/Target/MOS/MOSInstrCost.h
asiekierka: Fixed.

mysterymath (approved):

PR #326, opened 2023-08-13
Add 7 missing implicit quad mnemonics to 45GS02

mlund:
- [x] Add missing implicit mnemonics, e.g. `DEQ`, `INQ`, etc.
- [x] Rename `CPQ` to `CMPQ` to follow the MEGA65 book. The ACME `CPQ` syntax is is now an alias to `CMPQ` and thus still supported.
- [x] Reorder `all-45gs02opcodes.s` so that mnemonics are grouped by name. This makes it easier to compare with the MEGA65 book.
- [x] Remove unsupported mnemonics that was present in ACME but their use is currently undefined.
- [x] (Dis)assemble roundtrip OK


asiekierka:
It might be worth considering aliasing `CPQ` to `CMPQ`. Not sure, though.

mlund:
Alias added.

llvm/lib/Target/MOS/MOSInstrInfo.td
```diff
@@ -500,6 +500,15 @@ let Predicates = [Has45GS02] in {
 
 let DecoderNamespace = "45gs02" in {
 
+/// Q seudo register implicit (NEG NEG predix, 3 bytes)
```
mysterymath, line 503: nit: "pseudo", "prefix"

mysterymath (approved):

llvm/lib/Target/MOS/MOSInstrInfo.td
mlund: Fixed!

PR #333, opened 2023-08-16
MOS legalizer for floating-point operations

rweather:
Here is a first pass at adding floating-point support to LLVM-MOS.  Most of it was cut-and-pasted from the ARM legalizer.

Some intrinsics aren't working yet like G_INTRINSIC_ROUNDEVEN, so this is still a work in progress.  I'm putting it up for early review.

"float" has been confirmed working with my Mandelbrot example - I have some issues with the "double" version of Mandelbrot so I may have made a mistake somewhere.  I am investigating.

It is necessary to compile and install https://github.com/rweather/berkeley-softfloat-3-6502 into your LLVM-MOS-SDK install directory to get the softfloat code.  See that repo's README for instructions.  Then build a program with:

```
mos-sim-clang -Os -o program program.c -lbsoftfloat
```

If the `-lbsoftfloat` is omitted, then a lot of linker errors will occur for the missing intrinsics.  And there is no -lm or printf support yet, so only basic arithmetic is possible.

### DIFF
--- a/llvm/lib/Target/MOS/MOSLegalizerInfo.cpp
+++ b/llvm/lib/Target/MOS/MOSLegalizerInfo.cpp
@@ -237,20 +237,38 @@ MOSLegalizerInfo::MOSLegalizerInfo(const MOSSubtarget &STI) {
 
   // Floating Point Operations
 
+  // TODO: G_FNEG and G_FABS?
+
   getActionDefinitionsBuilder({G_FADD,       G_FSUB,
                                G_FMUL,       G_FDIV,
-                               G_FMA,        G_FPOW,
-                               G_FREM,       G_FCOS,
+                               G_FMA,        G_FREM,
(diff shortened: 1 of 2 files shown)


rweather:
I figured out what the issue with doubles was.  Berkeley Softfloat defines the internal types as:

```
typedef struct { uint32_t v; } float32_t;
typedef struct { uint64_t v; } float64_t;
```

The `float32_t` type was being passed in registers, the same as an `int32_t`.  The `float64_t` type was being passed by pointer because of the size of the structure.  Changing the definitions to this as recommended in Berkeley's code fixes the parameter passing:

```
typedef float float32_t;
typedef double float64_t;
```

No change to llvm-mos required - the change is in the softfloat library itself.  64-bit Mandelbrot is all good now.

mysterymath:
Thanks for putting this together!

PR #344, opened 2023-09-09
Add predefined MOS macros.

### LINKED ISSUE #332: Predefined macros for CPU type
It would be handy to have a predefined macro in the compiler for the CPU type; e.g. `__mosw65c02__` when `-mcpu=mosw65c02` is supplied on the command-line.  This will particularly benefit inline and out of line assembly that can customize its behavior with CPU-specific instructions.  A general "`__mos__`" define for the entire family may also be useful.

asiekierka:
Fixes #332. Could not come up with anything better, sorry.


mysterymath (approved):
This seems reasonable enough. Thanks!

Issue #321, opened 2023-07-08
[65816] 16-bit register mode support


asiekierka:
Part of https://github.com/llvm-mos/llvm-mos/issues/32 .

The main challenge here is being clever about emitting REP/SEP instructions to switch, and deciding which register widths to support and how - while the 8-bit registers A and B are effectively subsets of their 16-bit counterpart, they are not as cheap to access as on most LLVM-supported architectures; in addition, the 8-bit registers X and Y cannot be treated as subsets of their 16-bit counterparts, as their high portions are reset when the 16-bit index register mode is disabled.

Note that a target which provides the 65816's 16-bit register modes, but not the 24-bit address space, is already present in llvm-mos-sdk as "rpc8e" (RedPower 2's RPC8/e "fantasy" computer), and may as such prove useful for testing.

asiekierka:
Some nomenclature, first. The 65816 models the 16-bit accumulator as three distinct registers:

* `A` is the 8-bit accumulator register as used on the 6502,
* `C` is the 16-bit variant of the accumulator register, taking place of `A` when the 16-bit accumulator mode is enabled.
* `B` is the high portion of `C`; the high 8 bits. Its value (for the most part) can only be accessed in 8-bit mode by using `XBA`, which swaps `A` with `B`.

The 16-bit accumulator mode is often referred to as `a16`, with the 8-bit mode referred to as `a8`. The index register modes are referred to as `xy16`/`xy8` or `i16`/`i8` - I shall stick to the former.

Next, research. The ports of LCC (from 1994) and TCC (somewhat more recent) to generate code for the 65816 use `xy16` only, while toggling between `a16` (default) and `a8` in the only place where it matters significantly - for loading and storing variables. Modeling `A` is significantly easier, because:

* When switching between 16-bit and 8-bit accumulator modes, `B` is not destroyed; `XY` have their high bits 8-15 forced to `$00`.
* When switching between 16-bit and 8-bit accumulator modes, only `A` is affected; `XY` can only be affected as a pair.

[...]

PR #346, opened 2023-09-10
Implement REP/SEP emitting on 65816 targets.

asiekierka:
Another small step towards proper 65816 support...

This adds support for using `REP`/`SEP` instead of `BIT` for clearing/setting the `V` flag on 65816. This saves one byte and up to one cycle per invocation, as well as removes the need for `__set_v`.

The PR also provides facilities for modeling any other processor flags on the 65816 using `REP`/`SEP`, should the project choose to model `M` and `X` flags that way.

As always, I'm not sure if I'm doing things correctly at all.


llvm/lib/Target/MOS/MOSInstrLogical.td
```diff
@@ -266,6 +266,24 @@ def BITImmHUC6280 : MOSLogicalInstr, PseudoInstExpansion<(BIT_Immediate imm8:$r)
...
 }
 
+//===---------------------------------------------------------------------===//
+// REP/SEP Instructions
+//===---------------------------------------------------------------------===//
+
+def REPImm : MOSLogicalInstr {
+  let Predicates = [HasW65816Or65EL02];
+
+  dag OutOperandList = (outs FlagP:$dst);
```
asiekierka, line 276: The idea is that a write to a one-bit `Flag` can be used for one-bit load/stores, while `P` can be declared used for bulk load/stores (for instance, an optimization pass which combines adjacent REP/SEP instructions, if need be). This can be improved upon later, of course.

llvm/lib/Target/MOS/MOSInstrLogical.td
```diff
@@ -266,6 +266,24 @@ def BITImmHUC6280 : MOSLogicalInstr, PseudoInstExpansion<(BIT_Immediate imm8:$r)
   dag InOperandList = (ins Ac:$l, imm8:$r);
 }
 
+//===---------------------------------------------------------------------===//
+// REP/SEP Instructions
+//===---------------------------------------------------------------------===//
+
+def REPImm : MOSLogicalInstr {
```
asiekierka, line 273: An open question is if it'd be better to model `REPImm`/`SEPImm` as an `LDPImm`, supporting `P`, the flags, as well as the flag pairs like `NZ`.

However, this would require a later lowering stage to split `LDPImm` into a `REP` and a `SEP` opcode, as such a logical opcode could both clear and set bits, while `REP` and `SEP can only clear *or* set respectively. Feedback welcome!

mysterymath: I think that decision would depend on whether we it would remove  any duplicated case specification in the backend; without diving into it, it sounds prima facie like it would do the opposite.

llvm/lib/Target/MOS/MOSInstrInfo.cpp
```diff
@@ -608,15 +608,16 @@ void MOSInstrInfo::copyPhysRegImpl(MachineIRBuilder &Builder, Register DestReg,
               MOS::CMPImm, {MOS::C},
               {getRegWithVal(Builder, SrcReg, MOS::GPRRegClass), INT64_C(1)});
         } else {
-          assert(DestReg == MOS::V);
```
mysterymath, line 611: copyPhysReg should only be used for lowering COPY operations; these in turn should only be generated for allocatable registers, which still include only C and V. Accordingly, it looks like much of the code in this patch is effectively dead.

It doesn't seem like a good idea to broaden Anyi1 storage to general regs given C and V, either. The code here is mostly to allow successful codegen in cases where the compiler needs to stash something that would otherwise be produced from and used in the carry or overflow registers. We only model those as allocatable registers because we need a real model of register pressure to schedule instructions; the copy code here is a sign that scheduling has failed, and the pressure remains super-critical.

asiekierka: For non-65816 architectures, this will still only, effectively, work for `C` and `V`. The question is that of modeling writes to the 65816's new flags going forward, the ones controlling the bit width of A and X/Y between 16 and 8-bit: `M` and `X`. I can't imagine a world where they're efficiently allocated without the compiler being aware of the instruction scheduling consequences of switching between 8- and 16-bit modes; and I can't imagine an efficient 65816 code generator which doesn't switch between these modes (it's fairly common to do so in the SNES assembly codebases I've established, but it has consequences for X/Y particularly: upon switching from 16-bit mode to 8-bit mode, their high bits get zeroed out).

mysterymath: I guess I'd like to see more of a design on how that would work before adding this kind of complexity to the register model then. We may need to do something really custom to place mode switches; it doesn't seem likely that we'll be able to get the behavior we want out of the existing instruction selector and scheduler without doing wonky things like making every instruction have a data dependency  on M and X.

asiekierka:
Closing. Should be brought back when https://github.com/llvm-mos/llvm-mos/issues/321 is more concrete.

Issue #205, opened 2022-07-16
Zero page pointers


mysterymath:
Even with a whole-program zero page allocation pass after codegen, opportunities to use the zero page may still be missed.

For example, a loop over a small array could be modified to use the zero page instead. However, once code has been generated, there wouldn't be any benefit to moving the array to the zero page; a 16-bit pointer is used, so the performance is the same either way.

To handle this case, the compiler would need the internal concept of an 8-bit zero page pointer. A zero page allocation pass could be written before code generation at the IR level to try to move `alloca`s and globals to the zero page, which would make their pointers 8-bit. 8-bit pointers have much nicer properties than 16-bit ones; in particular, they can be kept in an index register and accessed through `LDA $00,X` or `LDA $0000,Y`, which is about twice the speed of `LDA (rs),Y`.

8-bit pointers can also be exposed directly to user code as an extension to standard C. This would allow programmers to manually allocate variables to the zero page and manipulate 8-bit pointers to them.

PR #348, opened 2023-09-10
Zero page pointers

asiekierka:
Initial work on https://github.com/llvm-mos/llvm-mos/issues/205 :

* Adds support for marking variables as zero page-stored in C: `char __zeropage value`.
* Adds support for one-byte-wide zero page pointers: - `char __zeropage* ptr`.
* Fixes support for HuC6280's redirected zero page. Turns out it was kind of broken and inconsistent throughout the codebase; that is no longer the case.
* Notably, **does not** change the zero page allocation mechanics yet, beyond supporting the new method for marking variables as zero-page.

TODO:

- [x] Fix ~~all~~ the *known* bugs. ~~It's pretty easy to get the compiler to crash right now, using the zero page variables.~~
- [x] Handle HuC6280's redirected zero page for zero page pointer casts.
- [x] Add tests.


asiekierka:
Okay, I can't think of much else to do here anymore; though I'm sure you'll have many ideas...

llvm/lib/Target/MOS/MOSLegalizerInfo.cpp
```diff
@@ -1717,7 +1826,14 @@ bool MOSLegalizerInfo::selectIndirectAddressing(LegalizerHelper &Helper,
     } else if (Helper.getKnownBits()
                    ->getKnownBits(Offset)
                    .countMaxActiveBits() <= 8) {
-      Index = Builder.buildTrunc(S8, Offset).getReg(0);
+      LLT OffsetTy = MRI.getType(Offset);
+      if (OffsetTy.getSizeInBits() < 8) 
+        Index = Builder.buildZExt(S8, Offset).getReg(0);
```
mysterymath, line 1831: Isn't there something like a ZextOrTrunc for cases like this? I've seen one floating around somewhere; there may be a way to clean up similar cases here.

llvm/lib/Target/MOS/MOSLegalizerInfo.cpp
```diff
@@ -1681,6 +1757,39 @@ bool MOSLegalizerInfo::tryAbsoluteIndexedAddressing(LegalizerHelper &Helper,
...
+    auto ConstOffset =
+        getIConstantVRegValWithLookThrough(PtrAddAddr->getOffsetReg(), MRI);
+    if (ConstOffset) {
+      Offset += ConstOffset->Value.getSExtValue();
+      Addr = PtrAddAddr->getBaseReg();
+    }
+  }
+  
+  auto AddrP = Builder.buildPtrToInt(S, Addr).getReg(0);
+  unsigned Opcode = isa<GLoad>(MI) ? MOS::G_LOAD_ABS_IDX : MOS::G_STORE_ABS_IDX;
```
mysterymath, line 1783: I think this'll cause some problems, since 8-bit pointers should wrap at 256, while these do not. We will need new GISEL opcodes and new MIs (I think?) to maintain that constraint. Otherwise, subtracting e.g. 1 from a zp pointer may bread, as it's a wrapping addition of 255. That actually makes it really nice that zp,x wraps!

llvm/lib/Target/MOS/MOS.h
```diff
@@ -45,6 +45,26 @@ template <typename ContainerTy> inline auto mbb_reverse(ContainerTy &&C) {
...
+namespace MOS {
+
+/// An integer that identifies all of the supported MOS address spaces.
+enum AddressSpace {
+  MemoryAS,
+  ZeroPageAS,
+  NumAddrSpaces,
+};
+
+template <typename T> AddressSpace getAddressSpace(T *V) {
```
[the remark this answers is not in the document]
asiekierka: It's not *typical* - it's lifted from the AVR target, which is I think the only one which does something similar to what we're doing here right now.

llvm/lib/Target/MOS/MOSLegalizerInfo.cpp
mysterymath: Nope, I see you sign extend here; that takes care of it. Would you mind filing an issue to do this at least; this will prevent ZP indexed addressing.

mysterymath:
It should also be pretty easy to add LSR support for this, since 8-bit pointers already meet it's constraints. Should be a relatively simple change to isLegalAddressingMode. I'd say to either do that here or file a separate issue, since I think it would just be an optimization.

llvm/lib/Target/MOS/MOSLegalizerInfo.cpp
mysterymath: Actually, I think sign extending would still cause problems; 255 past a pointer to 0 should point at 255, but sign extending yields 0xffff.

asiekierka: Later, at `addImm()`, the offset is ANDed with `0xFF` to ensure it wraps at 255.

Currently, the generic opcode `G_LOAD_ABS_IDX` translates to the logical opcode `LDAAbsIdx` (or `LDXAbsIdx`, or `LDYAbsIdx), which translates to the machine opcode `LDA_ZeroPageX`, and is then relaxed to `LDA_AbsoluteX` if the absolute address is outside of zero page bounds. My code sets the absolute address in `G_LOAD_ABS_IDX` such that it always fits within zero page bounds, and thus always emits a ZP indexed opcode as a result.

I suppose now that the zero page address space will be properly supported, one would ideally create new GISEL opcodes and get rid of this relaxation pass entirely. However, I think that issue is best left separate.

llvm/test/CodeGen/MOS/legalizer.mir
```diff
@@ -2343,6 +2424,80 @@ body: |
...
+body: |
+  bb.0.entry:
+    ; CHECK-LABEL: name: load_zp_constant_ptradd_zext_8
+    ; CHECK: [[COPY:%[0-9]+]]:_(s8) = COPY $a
+    ; CHECK-NEXT: [[LOAD_ABS_IDX:%[0-9]+]]:_(s8) = G_LOAD_ABS_IDX 123, [[COPY]](s8) :: (load (s8))
+    ; CHECK-NEXT: RTS implicit [[LOAD_ABS_IDX]](s8)
+    %0:_(p1) = G_CONSTANT i8 123
+    %1:_(s8) = COPY $a
+    %2:_(s16) = G_ZEXT %1
+    %3:_(p1) = G_PTR_ADD %0, %2
```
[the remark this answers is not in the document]
asiekierka: It's an omission when adapting the `p0` load/store tests to `p1`; likewise the one below.

llvm/lib/Target/MOS/MOSLegalizerInfo.cpp
mysterymath: > Currently, the generic opcode `G_LOAD_ABS_IDX` translates to the logical opcode `LDAAbsIdx` (or `LDXAbsIdx`, or `LDYAbsIdx`), which translates to the machine opcode `LDA_ZeroPageX`, and is then relaxed to `LDA_AbsoluteX` if the absolute address is outside of zero page bounds.

The logic in MOSMCInstLower that handles this calls `canUseZeroPageIdx`, which bails unless the address operand is a global value in the ZP address space. AFAICT, this change is the first time we rely upon the 8-bit wrapping behavior of LDA_ZeroPageX, so we'll need some kind of additional logic to ensure it.

asiekierka: Oh, right. I missed that.

I do agree that writing a new generic and logical opcode set here is the best option; however, this will take some time. I don't see any reason to cut corners here, though.

llvm/lib/Target/MOS/MOSInstructionSelector.cpp
```diff
@@ -1230,10 +1230,18 @@ bool MOSInstructionSelector::selectAddr(MachineInstr &MI) {
   MachineOperand Op = MI.getOperand(1);
   if (Op.isCImm())
     Op.ChangeToImmediate(Op.getCImm()->getSExtValue());
-  auto Instr =
-      Builder
-          .buildInstr(MOS::LDImm16, {MI.getOperand(0), &MOS::GPRRegClass}, {})
-          .add(Op);
+
+  LLT DestType = Builder.getMRI()->getType(MI.getOperand(0).getReg());
```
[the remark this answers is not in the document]
asiekierka: Is `legalizer.mir:select_zp` insufficient?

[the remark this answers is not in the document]
asiekierka: Fixed.

llvm/test/CodeGen/MOS/legalizer.mir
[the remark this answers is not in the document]
asiekierka: Fixed.

llvm/lib/Target/MOS/MOSLegalizerInfo.cpp
asiekierka: This will take a while - given `selectRMW` is about to grow from two to four addressing modes if zero-page accesses are to be always handled correctly, and six going forward with the 65816, I'd like to think if there's a better way to express `mi_match` sequences than copy-pasting for each addressing mode.

llvm/lib/Target/MOS/MOS.h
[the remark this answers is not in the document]
asiekierka: This should be resolved now.

llvm/lib/Target/MOS/MOSLegalizerInfo.cpp
asiekierka: Committed the "easier" proposal, in the form of adding a new MO_ZEROPAGE target flag. If you'd like me to do a full set of generic instructions/MIs, I'm open to that, but I'm not certain that's necessary.

asiekierka: That's not enough. Register allocation comes into play: an `abs,x` and `abs,y` mode exists for some opcodes, but no `zp,y`.

As such, for those opcodes which have distinct register allocation for `AbsIdx` and `ZpIdx`, separate logical opcodes have been defines; for the remaining opcodes, a common `Idx` suffix was adopted instead, to avoid confusion.

asiekierka:
I'm going to vote for adding LSR support separately; even if it's a simple change, this PR has already far outgrown its scope and become a bit unwieldy in size.

mysterymath (approved):
Alright, as far as I can tell, this LGTM. The change is getting pretty unwieldy, so it seems prudent to get it in; we'll likely have to iron out any remaining problems (if any) as they appear.

PR #352, opened 2023-09-17
MOS: Use upstream Motorola-style integer support in AsmLexer.

asiekierka:
With the addition of the M68K target, upstream has introduced support for Motorola-style integers, which cover parsing of dollar-prefixed hexademical numbers. This ever-so-slightly reduces our patch surface on top of upstream LLVM.

From an user perspective, this additionally adds support for percent-prefixed binary numbers - "#%10101010", for example. This is supported by ca65, 64tass, and probably other 6502 assemblers.

### DIFF
--- a/llvm/lib/Target/MOS/MCTargetDesc/MOSMCAsmInfo.cpp
+++ b/llvm/lib/Target/MOS/MCTargetDesc/MOSMCAsmInfo.cpp
@@ -23,7 +23,7 @@ MOSMCAsmInfo::MOSMCAsmInfo(const Triple &TT, const MCTargetOptions &Options) {
   CalleeSaveStackSlotSize = 0;
   SeparatorString = "\n";
   CommentString = ";";
-  DollarIsHexPrefix = true;
+  UseMotorolaIntegers = true;
   // Maximum instruction length across all supported subtargets.
   MaxInstLength = 7;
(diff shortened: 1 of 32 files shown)


llvm/lib/Target/MOS/MCTargetDesc/MOSMCAsmInfo.cpp
```diff
@@ -23,7 +23,7 @@ MOSMCAsmInfo::MOSMCAsmInfo(const Triple &TT, const MCTargetOptions &Options) {
   CalleeSaveStackSlotSize = 0;
   SeparatorString = "\n";
   CommentString = ";";
-  DollarIsHexPrefix = true;
+  UseMotorolaIntegers = true;
```
mysterymath, line 26: Hmm, if we specify this as part of the target, why do we need to specify and additional flag for the various tests? Is there a way we can make this setting stickier?

asiekierka: I followed upstream's lead here; it is done this way on the M68k target tests, too, and it did not want to work otherwise. I'm not too sure as to why this is the case myself...

mysterymath (approved):

johnwbyrd:
Oh nice, I missed that this was done.

Issue #355, opened 2023-09-20
Unrecognized opcodes in module level assembly


mlund:
Using non-6502 in module level assembler code I get the following compilation error. Inline and stand-alone assembly is OK.
```console
error: instruction requires a CPU feature not currently enabled
  ldz #0
```
when using e.g. `mos-mega65-clang++ -mcpu=mos45gs02 asm.cc` and the snippet shown below.
```cpp
// asm.cc
asm (
 ".text\n"
 ".global func\n"
 "func:\n"
 "ldz #0\n" // not ok
 "rts"
);
void func();

int main() {
  asm volatile ("ldz #0"); // ok
}

mlund:
This does seem to compile in Compiler Explorer. Not sure if this is related to my setup (git checkout on 15/9).

mysterymath:
Did a little digging on this, it looks like this may actually be a LTO bug upstream. They're not instantiating the target CPU when scanning the inline assembly for symbols before LTO occurs: 

https://github.com/llvm/llvm-project/blob/eb38edc82850c34bbbda7f7ce2235ec882b7d398/llvm/lib/Object/ModuleSymbolTable.cpp#L91

I'll start looking for an existing issue; if I can't find one, I'll try to make a X86 reproducer and file it upstream.

mysterymath:
Filed https://github.com/llvm/llvm-project/issues/67698 for this upstream.

mysterymath:
So, I did some digging into this upstream, and it looks like this error is partially spurious; the error is emitted, but the compile and link actually does eventually succeed. The only actual effects of this bug, other than the error message, is that module-level inline assembly is no longer scanned for symbol references to feed into the LTO symbol resolution table. This is already the situation for regular inline assembly, so it isn't a tremendous loss.

It's in a bit of a sticky situation upstream; a lot of code tends to assume that you can scan module-level inline assembly for symbols without knowing the CPU and feature set, and there are tools that operate on bitcode that couldn't possibly have that information, too.

mysterymath:
Taking this down to p2; it doesn't look like it's likely to be hard blocker on anything, just surprising and annoying, and it's architecturally quite difficult to fix.

Issue #354, opened 2023-09-20
Invalid addressing `stx <absolute>, y` accepted by assembler


cogwheel:
I unwittingly wrote some invalid code of the form:

```asm
stx some_buf, y
```

This is only valid if `some_buf` is on the zeropage. But the symbol was not marked as zeropage in the assembly file.

Here's an [example on compiler explorer](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:7,endLineNumber:10,positionColumn:7,positionLineNumber:10,selectionStartColumn:7,selectionStartLineNumber:10,startColumn:7,startLineNumber:10),source:'extern+char+buf%5B%5D%3B%0A%0Avoid+foo()%3B%0A%0Aasm+(%22%5C%0A.text+%5Cn%5C%0A.globl+foo+%5Cn%5C%0Afoo:+%5Cn%5C%0A++ldy+%231+%5Cn%5C%0A++stx+buf,+y+%5Cn%5C%0A++rts+%5Cn%5C%0A%22)%3B%0A%0Aint+main()+%7B%0A++++foo()%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:mos-nes-mmc3-trunk,deviceViewOpen:'1',filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'-O3',overrides:!(),selection:(endColumn:12,endLineNumber:10,positionColumn:12,positionLineNumber:10,selectionStartColumn:12,selectionStartLineNumber:10,startColumn:12,startLineNumber:10),source:1),l:'5',n:'0',o:'+llvm-mos+nes-mmc3+(Editor+%231)',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)

The resulting machine code for this instruction is `96 00`, aka `stx $00, y`

johnwbyrd:
[...]

PR #357, opened 2023-09-23
SPC700 support

asiekierka:
## Introduction

The SPC700 is a community mnemonic for a Sony-designed 6502-like CPU. More information is available [on the SNESdev Wiki](https://snes.nesdev.org/wiki/SPC-700_instruction_set).

In this case, "6502-like" means that the CPU incorporates most of the instruction set of the 6502, as well as some extensions from the 65C02 family, and some custom extensions on top; however, it is not fully compatible with the 6502. While most of the incompatibilies (missing decimal mode, altered processor flag layout, swapped meaning of interrupt bit) don't matter for the purposes of the LLVM backend, two very much do:

1. The instruction set is allocated completely differently, with each opcode (except Bxx conditional branches) in a different spot compared to its 65C02 counterpart,
2. The official assembler syntax is likewise completely different, with most opcodes renamed and different syntax for addressing modes on top.

## Implementation

With regards to point 1, a new TableGen file is defined, containing all of the SPC700 opcode definitions.

With regards to point 2, a decision has been made to support both the official syntax and a community-derived set of 6502-style mappings (as aliases to the official syntax).

In addition, **the following changes will need to be made to the ELF specification**:

* The addition of a mutually exclusive `EM_MOS_SPC700` (`0x20000`) machine type bitflag,
* The addition of a `R_MOS_ADDR13` relocation type.

## TODO

- [x] Write an ~~MCInst lowering pass~~ MOSASMPrinter hook which translates logical instructions into SPC700-specific machine instructions.
- [x] Provide workarounds for missing opcodes (such as BIT or JMP indirect) at the compiler level, where necessary.

## Future work

There are many optimizations which can be done with regards to the newly provided opcodes, but these are out of scope for this PR (and will probably be delayed until zero page pointers are dealt with regardless).


asiekierka:
With regards to implementing backend support, there's prior art available at https://github.com/jackoalan/llvm-mos/commits/spc700-build which I only learned of *after* I finished this PR.

asiekierka:
This is now ready for review, though the following three issues should probably be closed prior to merging:

* https://github.com/llvm-mos/llvm-mos/issues/358
* https://github.com/llvm-mos/llvm-mos/issues/355 - if my hunch is right, this will prevent writing SPC700 code *at all* in module-level assembly,
* https://github.com/llvm-mos/llvm-mos/issues/354 - this extends to all invalid 6502 opcodes in SPC700 mode.

johnwbyrd:
I have glanced through the above changes; although I haven't run it yet, I feel that generally you have taken the correct strategy wrt SPC700.  Thank you in particular for all the test cases.

llvm/lib/Target/MOS/MOSInstrInfo.cpp
```diff
@@ -610,12 +610,20 @@ void MOSInstrInfo::copyPhysRegImpl(MachineIRBuilder &Builder, Register DestReg,
...
+              STI.hasGPRStackRegs() ? MOS::GPRRegClass : MOS::AcRegClass;
 
           if (StackRegClass.contains(SrcReg)) {
-            Builder.buildInstr(STI.has65C02() ? MOS::PH_CMOS : MOS::PH, {}, {SrcReg});
-            Builder.buildInstr(STI.has65C02() ? MOS::PL_CMOS : MOS::PL, {SrcReg}, {})
-                .addDef(MOS::NZ, RegState::Implicit);
+            if (STI.hasSPC700()) {
+              // SPC700 pops do not set NZ flags, so use CMP instead.
+              // TODO: Investigate doing the same for MOS.
+              Builder.buildInstr(MOS::CMPImm, {MOS::C}, {SrcReg, INT64_C(0)})
```
mysterymath, line 619: The copy operations that this expands aren't recorded as clobbring C, so C may be live across them, and cannot safely be clobbered. We'll need to find some other way to do this.

llvm/lib/Target/MOS/MOSInstrInfo.cpp
```diff
@@ -1017,6 +1027,21 @@ void MOSInstrInfo::expandLDImm1(MachineIRBuilder &Builder) const {
     break;
   case MOS::V:
     if (Val) {
+      if (STI.hasSPC700()) {
+        // SPC700 does not have BIT, so we use stack operations to specifically
+        // set V.
+        Builder.buildInstr(MOS::PH, {}, {Register(MOS::A)});
```
mysterymath, line 1033: Instead of pushing A, you can create a virtual register of the Ac register class. This allows the register scavenger to later ensure A is free; it does so globally, so it can optimize adjacent save/restore logic.

llvm/lib/Target/MOS/MOSInstructionSelector.cpp
```diff
@@ -1890,6 +1902,21 @@ bool MOSInstructionSelector::selectUnMergeValues(MachineInstr &MI) {
...
 }
 
+bool MOSInstructionSelector::selectBrIndirect(MachineInstr &MI) {
+  if (STI.hasSPC700()) {
+    // SPC700 indirect jumps are indexed by X. Since G_BRINDIRECT does not
+    // directly support indexed effective addresses, we simply make X zero.
+    // TODO: Combiner could detect indexed indirect jump addresses.
+    MachineIRBuilder Builder(MI);
+    Register XZero =
+        Builder.buildInstr(MOS::LDImm, {MOS::X}, {INT64_C(0)}).getReg(0);
```
mysterymath, line 1912: This early, the destination should be a virtual register of Xc class, not a physical register.

llvm/lib/Target/MOS/MOSInstrInfo.cpp
asiekierka: Proposal: For all platforms, use `ORA #0` for A, and `DEX/INX` + `DEY/INY` for X/Y. Both only modify N and Z, and both end in the register's value being preserved (the register's value is unchanged). At the same time, they're much faster, not involving memory accesses: a `PH`/`PL` pair is 2 bytes, 7 cycles; `ORA #0` is 2 bytes, 2 cycles; a `DE/IN` pair is 2 bytes, 4 cycles.

llvm/lib/Target/MOS/MOSInstructionSelector.cpp
asiekierka: Should be fixed.

llvm/lib/Target/MOS/MOSInstrInfo.cpp
asiekierka: Should be fixed, though I'm not sure if I did it correctly.

jackoalan:
This is incredible @asiekierka! Thank you for continuing my SPC700 work. I’ve been short on time to work on llvm-mos, but I’m pleased to see the level of community engagement it’s gotten.

asiekierka:
@jackoalan Thank you for the kind words! 

For SPC700/65C816 simulation, my plan is to adapt Ares's emulation cores into a separate simulator (which will also cover some other platforms I want to target in the future). These are known to be quite accurate, as well as permissively licensed (MIT).

llvm/lib/Target/MOS/MOSInstrInfo.cpp
mysterymath: Ah, that's much better! Why didn't I think of that?

mysterymath (approved):

llvm/lib/Target/MOS/MOSInstrInfo.cpp
asiekierka: It only occured to me when going back to the list of "opcodes which update N and Z without updating other flags", to be fair. I would probably have done the same as you.

PR #374, opened 2023-10-01
MOS: Reduce 65C02 lowering code complexity by varying DE/IN/PH/PL log…

asiekierka:
…ical opcode choice based on register classes.

This strategy appears to be present in LLVM - the X86 target uses it to vary supported register classes between 32-bit and 64-bit mode. One important note from testing and studying other targets' use of AltOrder is that there can be multiple AltOrders, and the "main" order doesn't have to ever be returned, but the "main" order must be a superset of all AltOrders.

This PR uses it to clean up 65C02 logical opcode handling, but I foresee it being of potential greater use for two other subtargets:

* 65C816: according to my current plan for generating native code on 65C816, the X and Y register will be used almost exclusively in an 16-bit form, while the A register will be used in both its 8- and 16-bit form.
* SPC700: this target adds a few new addressing modes to arithmetic, logical, and load/store instructions, which among other things allow bumping the register class for many opcodes from "Ac" to "AImag8".

### DIFF
--- a/llvm/lib/Target/MOS/MOSFrameLowering.cpp
+++ b/llvm/lib/Target/MOS/MOSFrameLowering.cpp
@@ -118,7 +118,7 @@ bool MOSFrameLowering::spillCalleeSavedRegisters(
       continue;
     if (!StackRegClass.contains(Reg))
       Reg = Builder.buildCopy(&StackRegClass, Reg).getReg(0);
-    Builder.buildInstr(STI.hasGPRStackRegs() ? MOS::PH_CMOS : MOS::PH, {}, {Reg});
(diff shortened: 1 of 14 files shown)


mysterymath:
This looks suspicious as a long-term strategy, since as far as I can tell, the alternate orders aren't taken into account when scheduling for register pressure. Accordingly, this would tend to cause the scheduler to make many instances of e.g. A live at once thinking that AImag8 , but the register allocator would only later discover that its other choices are actually reserved.

asiekierka:
That's the information I lacked; thank you.

Ah well. I'm worried the amount of alternate loglcal instructions will grow to scary amounts, though, but if there's no other way...

Issue #239, opened 2022-10-15
Convert ca65 object files to llvm-mos ELF in llvm-objcopy.


mysterymath:
Having a build step available to convert ca65 object files to ELF object files would make it relatively straightforward to incorporate existing ca65 assembly files into an llvm-mos link. This would allow existing assembly "libraries" (typically standalone files) that make heavy use of ca65 features to be used with our SDK's relatively sophisticated linker system.

chorman0773:
I'd note that some things will not be convertible. xo65 objects have far
more complex relocation options available than elf. I discovered this in
attempting to attach to my existing binary format structures.

mysterymath:
This is specifically for the intermediate object files used by ca65 and ld65, not the o65 load-time relocatable object format. We also have considerable latitude to set up any additional relocations necessary to make this work, since our ELF extension is well-defined, but project-specific.

chorman0773:
Yes, xo65 refers to that format use by the cc65 project in general. o65 is
closer to a.out and is quite well-behaved.

There are some issues that are fundamental to elf as well. xo65 allows, for
example, `foo+bar` as a relocation in the object file, where `foo` and
`bar` are both non-absolute symbols.

mysterymath:
Ah, I wasn't familiar with the nomenclature; references to that as a filename are very scant in their documentation.

I don't think this creates any intrinsic difficulty, since ELF allows you to compose relocations. A pair of relocations could add two relative symbols to the same base address.

chorman0773:
Can it mutiply or xor two symbols?

This is what I worked off of previously: https://github.com/Rapdorian/xo65

mysterymath:
Sigh, thanks for the link; I hadn't expected that they'd put an entire expression grammar into their object file format.
I don't think it's fatal, but it will require some hackery. Off the top of my head, it may be possible to create a generalized ca65 expression grammar relocation that uses an auxiliary symbol containing the expression. But I'd need to get into the linker and see what's feasible.

mysterymath:
I've spent some time working on this, and I think I have a strategy for making this a bit more hands-off.

[...]

PR #395, opened 2023-10-28
Wrap od65 and ld65 to link xo65 object files

mysterymath:
This PR adds a complete subsystem to LLD for handling xo65 object files (ca65's output format) in the course of a normal link.

When input files are being parsed, xo65 files are detected via their magic number and `od65` is invoked to parse them, either using the system path or the `--od65-path` flag. This allows the files to participate in normal symbol resolution and for xo65 segments to be placed by the linker.

Afterwards, `ld65` is invoked (PATH or `--ld65-path`) to relocate the xo65 object files and and collect the resulting binary contents. The map file is also examined to determine the final values of xo65 symbols, which are then used to relocate ELF files.

While there is much xo65-specific logic added to LLD, the general "linker enclave" approach is not very xo65-specific, and it could be broken out into an effective interface. I've not done so here, for want of a second implementation of that interface to drive its design.

ELF section names, types, and flags, are much richer than xo65's segments, so this information is encoded into xo65 segment names via an underscore-based escaping scheme:
```
  __  -> _             (underscore)
  _d  -> $             (dollar)
  _h  -> -             (hyphen)
  _p  -> .             (period)
  _tn -> SHT_NOBITS    (@nobits)
  _tp -> SHT_NOBITS    (@progbits)
  _fw -> SHF_WRITE     (w)
  _fx -> SHF_EXECINSTR (x)
```
Finally, this does not work with relocatable (-r) links, and this produces a link error.


mysterymath:
@jroweboy @asiekierka

mysterymath:
See llvm-mos/llvm-mos-sdk#234

mysterymath:
Fixes #239

Issue #233, opened 2022-10-12
Port relevant parts of compiler-rt, llvm-libc, and libc++


mysterymath:
The dearth of standard library functionality in LLVM's SDK at the very least presents the image that the compiler is less far along than it actually is. Quite a bit of the standard library functionality in the SDK that is missing is also rather useful, e.g. the time functions.

Given that LLVM itself has set a goal to provide good pure C versions of all of these routines, through its compiler-rt, llvm-libc, and libc++ subprojects, we should at least attempt to port these to llvm-mos. Whatever we can get working well for the 6502 can be imported into the SDK during its build process via the exsting library merging abstraction.

Given that upstreaming llvm-mos is a distant glimmer, we can make changes to these routines in LLVM to make them suitable for the 6502. This can be guarded by macro flags in the usual way.

mysterymath:
Split into #420, #421, and #422

PR #418, opened 2023-11-22
Build compiler-rt builtins

mysterymath:
This adds a build of compiler-rt's builtins (libclang_rt.a), a libgcc replacement. We disable crt0 and integer division libcalls, since the versions already in the SDK are better tuned for the 6502. We gain mostly encyclopedic coverage of libgcc and a license unencumbered soft float library, albeit one also poorly optimized for the 6502.


mysterymath:
See issue #233

PR #419, opened 2023-11-28
Strict inlining costs

mysterymath:
The inliner will apply negative boosts to inlining costs if runtime
desirable situations occur. In particular, inlining is strongly
preferred if an indirect call can be folded away as a result.
Unfortunatley, this can blow up sizes on the 6502 arbitrarily, and
worse, it affects our printf implementation.

This change adds a new TTI hook, strictInliningCosts(), that prevents
this boost from altering the true costs. Size is paramount for our
target, so we should disable other boosts using this flag as they cause
problems.

### DIFF
--- a/llvm/lib/Target/MOS/MOSTargetTransformInfo.h
+++ b/llvm/lib/Target/MOS/MOSTargetTransformInfo.h
@@ -60,6 +60,8 @@ public:
   bool isValidAddrSpaceCast(unsigned FromAS, unsigned ToAS) const {
     return true;
   }
+
+  bool strictInliningCosts() const { return true; }
 };
 
 } // end namespace llvm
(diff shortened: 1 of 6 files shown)


llvm/lib/Target/MOS/MOSTargetTransformInfo.h
```diff
@@ -60,6 +60,8 @@ class MOSTTIImpl : public BasicTTIImplBase<MOSTTIImpl> {
   bool isValidAddrSpaceCast(unsigned FromAS, unsigned ToAS) const {
     return true;
   }
+
+  bool strictInliningCosts() const { return true; }
```
asiekierka, line 64: Out of curiosity: Wouldn't it be a good idea to disable this hook for something like `-O3`?

mysterymath: Hmm, yeah, that seems appropriate.

PR #448, opened 2024-02-24
MOS: SPC700: Emit MOV imag8, immediate instructions; refactor aspects of register class-based instruction variations.

asiekierka:
As with every logical instruction variant added, this involves a slightly annoying set of subtarget if-checks. However, being able to store an immediate in an imaginary register directly should be rather useful.


asiekierka:
Before this is merged, though, I'd like to step back for a second and consider if there's a better way of doing this. This is because the SPC700 extends a lot of opcodes in this manner. In addition to loads/stores, ADC, SBC, CMP, AND, OR, EOR all now receive `Imag8, Imag8` and `Imag8, i8imm` variants, thus widening their register classes from `Ac, imm8` and `Ac, Imag8` (or GPR for `CMP`) to `AImag8, imm8` and `AImag8, Imag8` respectively.

llvm/lib/Target/MOS/MOSInstructionSelector.cpp
```diff
@@ -1337,11 +1343,12 @@ std::pair<Register, Register>
 MOSInstructionSelector::selectAddrLoHi(MachineInstr &MI) {
   MachineIRBuilder Builder(MI);
   LLT S8 = LLT::scalar(8);
-  auto LoImm = Builder.buildInstr(MOS::LDImm, {S8}, {}).add(MI.getOperand(1));
+  auto LdImm = STI.hasSPC700() ? MOS::LDImmSPC700 : MOS::LDImm;
+  auto LoImm = Builder.buildInstr(LdImm, {S8}, {}).add(MI.getOperand(1));
```
johnwbyrd, line 1347: Is there a reason these LdImm constants have been taken out of the MOS namespace?

johnwbyrd, line 1347: Is there a reason that the LdImm constant was taken out of the MOS namespace?

llvm/lib/Target/MOS/MOSInstructionSelector.cpp
```diff
@@ -1337,11 +1343,12 @@ std::pair<Register, Register>
...
   LoImm->getOperand(1).setTargetFlags(MOS::MO_LO);
   if (!constrainSelectedInstRegOperands(*LoImm, TII, TRI, RBI))
     llvm_unreachable("Cannot constrain instruction.");
-  auto HiImm = Builder.buildInstr(MOS::LDImm, {S8}, {}).add(MI.getOperand(1));
+  auto HiImm = Builder.buildInstr(LdImm, {S8}, {}).add(MI.getOperand(1));
```
johnwbyrd, line 1351: Same question as above.

llvm/lib/Target/MOS/MOSInstrInfo.h
```diff
@@ -131,6 +131,7 @@ class MOSInstrInfo : public MOSGenInstrInfo {
   void expandLDIdx(MachineIRBuilder &Builder, bool ZP) const;
   void expandLDImm1(MachineIRBuilder &Builder) const;
   void expandLDImm16(MachineIRBuilder &Builder) const;
+  void expandLDImm16SPC700(MachineIRBuilder &Builder) const;
```
johnwbyrd, line 134: I don't question the implementation concept, but I am concerned that the specific naming that you've chosen to burn into the API might become invalid if a future processor elects to use the same addressing mode.  Consider renaming to LDImm16Imag8Imm, or something that you consider to be more descriptive and less processor specific.  Hopefully changing this would be a single editor operation and not too painful.

llvm/lib/Target/MOS/MOSMCInstLower.cpp
```diff
@@ -382,15 +382,30 @@ void MOSMCInstLower::lower(const MachineInstr *MI, MCInst &OutMI) {
     return;
   }
   case MOS::LDImm:
+  case MOS::LDImmSPC700:
   case MOS::LDAbs:
   case MOS::LDImag8:
   case MOS::STAbs: {
+    if (MI->getOpcode() == MOS::LDImmSPC700) {
```
johnwbyrd, line 389: As above.  Consider divorcing the particular relocation from that specific processor (in name only), thus making it a little more generic.

llvm/test/CodeGen/MOS/instruction-selector-ir-spc700.mir
```diff
@@ -1,22 +1,147 @@
...
+    ; CHECK: [[LDImm16SPC700_:%[0-9]+]]:imag16 = LDImm16SPC700 blockaddress(@br_indirect, %ir-block.entry)
     ; CHECK-NEXT: [[LDImm:%[0-9]+]]:xc = LDImm 0
-    ; CHECK-NEXT: JMPIndir %0, implicit killed [[LDImm]]
+    ; CHECK-NEXT: JMPIndir [[LDImm16SPC700_]], implicit killed [[LDImm]]
     ; CHECK-NEXT: RTS
     %0:any(p0) = G_BLOCK_ADDR blockaddress(@br_indirect, %ir-block.entry)
     G_BRINDIRECT %0
     RTS
 ...
+---
```
johnwbyrd, line 52: Thank you for creating this detailed test suite, I'll have to let @mysterymath comment on it.

llvm/test/CodeGen/MOS/postrapseudos-spc700.mir
```diff
@@ -40,3 +40,12 @@ body: |
     $y = LDAbsIdx 1234, $x
     $c = CMPImm $a, 0
 ...
+---
```
johnwbyrd, line 43: Thank you for your always clean and well-considered change requests!

johnwbyrd (changes requested):
Thank you for your always clean and well considered change requests.  Comments inline.  If you strongly disagree, we could go ahead.

llvm/lib/Target/MOS/MOSInstrInfo.cpp
```diff
@@ -1118,6 +1121,35 @@ void MOSInstrInfo::expandLDImm16(MachineIRBuilder &Builder) const {
   MI.eraseFromParent();
 }
 
+void MOSInstrInfo::expandLDImm16SPC700(MachineIRBuilder &Builder) const {
```
mysterymath, line 1124: This is intrinsically similar enough to LDImm16 that it should be more maintainable to have one implementation that switches out the differences by examining the opcode.

llvm/lib/Target/MOS/MOSInstructionSelector.cpp
```diff
@@ -1337,11 +1343,12 @@ std::pair<Register, Register>
 MOSInstructionSelector::selectAddrLoHi(MachineInstr &MI) {
   MachineIRBuilder Builder(MI);
   LLT S8 = LLT::scalar(8);
-  auto LoImm = Builder.buildInstr(MOS::LDImm, {S8}, {}).add(MI.getOperand(1));
+  auto LdImm = STI.hasSPC700() ? MOS::LDImmSPC700 : MOS::LDImm;
```
mysterymath, line 1346: It seems like a buildLDImm abstraction might pull a lot of weight in the instruction selector; it should be possible to factor out the SPC700 vs others logic into one place. The difference doesn't affect the logical arguments of the LDImm16, only the scratch register.

llvm/test/CodeGen/MOS/instruction-selector-ir-spc700.mir
mysterymath: LGTM

llvm/lib/Target/MOS/MOSInstructionSelector.cpp
asiekierka: It hasn't been taken out; I use a temporary variable to select between `LDImmSPC700` and `LDImm` to keep the code a little bit more readable.

llvm/lib/Target/MOS/MOSInstrInfo.h
asiekierka: I don't agree with the rationale presented.

First of all, if I recall my LLVM correctly, this compile stage and thus naming is internal to the compiler, and thus can be broken by us at any time without causing a breaking change to the end user. This means that, if such a future CPU appears, this name can easily be changed.

Second of all, there is precedent for logical and pseudo instructions which are specific to a given CPU to be suffixed with that CPU's name or characteristics - f.e. `PH_CMOS` and `PL_CMOS` for 65C02-specific variants of `PH` and `PL`. It helps highlight a code path as "specific to only some 6502s", rather than "available to all 6502s".

However, I do not have a strong opinion here.

llvm/lib/Target/MOS/MOSInstrInfo.cpp
asiekierka: I was conflicted on that, but have no strong opinion.

llvm/lib/Target/MOS/MOSInstructionSelector.cpp
asiekierka: `buildLDImm` and `buildLDImm16`, you mean?

llvm/test/CodeGen/MOS/postrapseudos-spc700.mir
asiekierka: Thank you for the kind words.

llvm/lib/Target/MOS/MOSInstructionSelector.cpp
mysterymath: > `buildLDImm` and `buildLDImm16`, you mean? As each only sees one use in `MOSInstructionSelector`.

Ah, I missed the split there.

> Perhaps it's time to add a new helper file, which handles the subtarget-specific logic for building a given instruction, and refactoring away existing cases (like PH vs PH_CMOS, IN vs IN_CMOS, Inc vs IncCMOS) in that general direction?

That's part of what I've seen InstrInfo used in other target, IIRC. General helpers that use the subtarget info to build specific versions of broader logical instructions. It made sense to have all of that switching inline here when it was relatively small, but we should definitely move it out if it starts making the higher level logic here difficult to read. It could also still live in this file, probably, just out-lined.

llvm/lib/Target/MOS/MOSInstrInfo.h
johnwbyrd: I suppose I can risk the chance that someone will decide to implement a variant of the SPC700.

mysterymath (approved):
LGTM. Would you like to try to break out a `build` helper suite before I merge this, or should we leave exploring that until later?

llvm/lib/Target/MOS/MOSInstructionSelector.cpp
asiekierka: I see no reason to rush this PR, so I've pushed a first draft of a `MOSInstrBuilder` helper header, in line with what a few other targets do, however it's probably not idiomatic - it's mostly an extraction of stuff related to subtarget instruction selection that was previously scattered across the codebase.

asiekierka:
Having thought about it, I'd like to propose a solution to some concerns raised that should also scale well for the other operations: Having *both* an `LDImmSPC700` and an `LDImag8Imm`. In this case, the `LDImmSPC700` would act solely as a register allocator pseudo that is later expanded in post-RA into either `LDImm` (if the register being loaded into is A, X, or Y) or `LDImag8Imm` (if it's one of the imaginary registers).

The benefit here is that:

* After RA, both are treated (correctly) as distinct instructions with distinct semantics. Passes can only care about one of them without having to worry about the other, avoiding being edited whenever a new subarchitecture comes up with a new register.
* The final `MCInstLower` pass is simplified, with fewer edge cases.

@mysterymath Do you think this approach would work better/equal to the current one?

mysterymath:
> Having thought about it, I'd like to propose a solution to some concerns raised that should also scale well for the other operations: Having _both_ an `LDImmSPC700` and an `LDImag8Imm`. In this case, the `LDImmSPC700` would act solely as a register allocator pseudo that is later expanded in post-RA into either `LDImm` (if the register being loaded into is A, X, or Y) or `LDImag8Imm` (if it's one of the imaginary registers).
> 
> This is counter-intuitive, but I believe modeling _two_ instructions here will actually _simplify_ reasoning about the compiler backend. The benefit here is that:
> 
> * After RA, both are treated (correctly) as distinct instructions with distinct semantics. Passes can only care about one of them without having to worry about the other, avoiding being edited whenever a new subarchitecture comes up with a new register.
> * The final `MCInstLower` pass is simplified, with fewer edge cases.
> * We're going to have a lot more of these going forward - as suggested above, the 65CE02 and 65C816 both introduce new registers to the equation, while SPC700's additional addressing modes affect imaginary registers across six more opcodes. Even a small logic simplification is thus a big victory.
> 
> @mysterymath Do you think this approach would work better/equal to the current one?

Hmm, I don't hate it. We'd generally sought use logical instructions to regularize the instruction set, but the different targets each introduce different sets of instruction sets to regularize. The target specifications throw a wrench in thinking of it that way though: there are actually N different instructions set, and each would have its own independent regularization. I think your proposal (if I'm understanding it correctly; I may not be) is more along the lines of regularizing the joint instruction set of all targets simultaneously, which is actually what we should have been doing, since that's what we actually support in the backend.

llvm/lib/Target/MOS/MOSInstrBuilder.h
```diff
@@ -0,0 +1,78 @@
...
+#include "MOSSubtarget.h"
+#include "MCTargetDesc/MOSMCTargetDesc.h"
+#include "llvm/CodeGen/MachineFrameInfo.h"
+#include "llvm/CodeGen/MachineInstrBuilder.h"
+#include "llvm/CodeGen/MachineMemOperand.h"
+#include "llvm/CodeGen/MachineOperand.h"
+
+namespace llvm {
+
+static inline unsigned getIncOpcode(const MachineIRBuilder &Builder) {
```
mysterymath, line 26: I instantly like this a lot.

llvm/lib/Target/MOS/MOSInstructionSelector.cpp
asiekierka: Okay, after a bunch of Discord discussion I'm happy with the refactors here.

llvm/lib/Target/MOS/MOSMCInstLower.cpp
asiekierka: I found a way to divorce it altogether.

asiekierka:
After a bunch of conversations on Discord and here, I think this PR is in a good spot.

There's much more work to be done now that we have a fairly inexpensive way to vary register classes - the `Z` register on the 65CE02 and widening `A, Imag8` ALU instructions to `AImag8, Imag8` comes to mind - but I don't want to put everything in one unmergeable PR. A few steps at a time.

llvm/lib/Target/MOS/MOSInstrPseudos.td
```diff
@@ -149,10 +151,10 @@ def DEC : MOSPseudo {
...
 }
 
+// As the 65C02 allows incrementing any register with INC/DEC only,
+// separate pseudos are defined which do not define C or V.
+// These correspond directly to INC and DEC.
 let Predicates = [HasGPRIncDec] in {
-  // As the 65C02 allows incrementing any register with INC/DEC only,
-  // separate pseudos are defined which do not define C or V.
-
   def IncCMOS : MOSPseudo {
```
mysterymath, line 158: Do we still need these? It seems like we could directly select the logical instructions `INC` and `DEC` on targets that support the 65c02 opcodes. IIRC, we should be able to break out the tablegen patterns as `Pat` definitions  with an appropriate `Predicate`.

mysterymath (approved):

llvm/lib/Target/MOS/MOSInstrPseudos.td
asiekierka: The thought didn't occur to me, but the latest commit should do the trick.

mysterymath (approved):

PR #452, opened 2024-03-02
MOS: Cleanup copy cost calculation, add 65CE02 cost values.

asiekierka:
Recently, I discovered that the 65CE02 adds yet another collection of distinct instruction timings. With that (and the observation of potential maybe-mistakes-but-I'm-no-longer-sure in the cost calculation logic), I decided to also use this opportunity to do some refactoring.

### DIFF
--- a/llvm/lib/Target/MOS/MOSRegisterInfo.cpp
+++ b/llvm/lib/Target/MOS/MOSRegisterInfo.cpp
@@ -701,10 +701,22 @@ bool MOSRegisterInfo::getRegAllocationHints(Register VirtReg,
     return true;
   }
 
-  MOSInstrCost INCzp =
-      MOSInstrCost(2, STI.hasHUC6280() ? 6 : (STI.hasSPC700() ? 4 : 5));
-  MOSInstrCost ASLzp =
-      MOSInstrCost(2, STI.hasHUC6280() ? 6 : (STI.hasSPC700() ? 4 : 5));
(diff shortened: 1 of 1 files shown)


llvm/lib/Target/MOS/MOSRegisterInfo.cpp
```diff
@@ -991,18 +1006,18 @@ MOSInstrCost MOSRegisterInfo::copyCost(Register DestReg, Register SrcReg,
...
       // LDImm; BNE; LDImm;
-      MOSInstrCost Cost = MOSInstrCost(2, 2) * 2 + MOSInstrCost(2, 3);
+      MOSInstrCost Cost = LoadImmCost * 2 + BranchCost;
       if (Tmp != DestReg)
         Cost += copyCost(DestReg, Tmp, STI);
       return Cost;
     }
     if (STI.hasSPC700()) {
       // PHA, PHP, PLA, ORA #imm, PHA, PLP, PLA, BR, CLV
-      return MOSInstrCost(1, 4) * 6 + MOSInstrCost(2, 2) + MOSInstrCost(1, 3) +
```
asiekierka, line 1001: This is the contentious part (with a similar mistake repeated below). `BR` is 2 bytes and 3 cycles; `CLV` is 1 byte and 2 cycles.

mysterymath (approved):

Issue #383, opened 2023-10-12
Missed optimization opportunity when doing a commutative operation on A and X (or Y?)


juj:
Consider a function

```c++
#include <stdint.h>

uint8_t ADD_U8_SAT(uint8_t a, uint8_t b)
{
  uint8_t result = a + b;
  if (result < a) return 0xFF;
  return result;
}
```
It generates
```asm
; ADD_U8_SAT(unsigned char, unsigned char):                       ; @ADD_U8_SAT(unsigned char, unsigned char)
        sta     __rc2
        txa
        clc
        adc     __rc2
        bcc     .LBB0_2
        lda     #255
.LBB0_2:
        rts
```
(as seen on Godbolt at https://godbolt.org/z/qWc9s8enx )

It looks like there might be an inefficiency here with respect to how `A` and `X` are handled. Unless I made an oopsie in logic, the above code could be optimized into a more compact form:

```asm
; ADD_U8_SAT(unsigned char, unsigned char):                       ; @ADD_U8_SAT(unsigned char, unsigned char)
        stx     __rc2
        clc
        adc     __rc2
        bcc     .LBB0_2
        lda     #255
.LBB0_2:
        rts
```

A general optimization pattern here could be of form

```asm
; inputs A and X
STA  _some_address
TXA
COMMUTATIVE_OPERATION _some_address
; _some_address no longer in use after this line
```
could be transformed to
```asm
; inputs A and X
STX  _some_address
COMMUTATIVE_OPERATION _some_address
```
to win the `TXA` operation. Likewise instead of `X` one might have `Y` register at play.

Here `COMMUTATIVE_OPERATION` could be any of `AND`, `EOR`, `ORA` or `ADC`.

asiekierka:
The issue here is that the `TwoAddressInstructionPass` chooses to commute `A` and `X` for `ADCImag8`, which leads to this sub-par code generation. Disabling that pass generates the expected `STX` code.

There is a target hook, `TargetInstrInfo::hasCommutePreference`, and force-disabling commutation there is effective. The question is, for what cases should we be returning false here?

PR #453, opened 2024-03-03
MOS: Avoid commutating operands when doing so would lead to less trivial copies.

asiekierka:
Fixes https://github.com/llvm-mos/llvm-mos/issues/383 .

I can't help but wonder if this case couldn't be caught in a more generalized manner - but I lack the wisdom to get there myself :smiling_face_with_tear:

### DIFF
--- a/llvm/lib/Target/MOS/MOSInstrInfo.cpp
+++ b/llvm/lib/Target/MOS/MOSInstrInfo.cpp
@@ -267,6 +267,66 @@ bool MOSInstrInfo::findCommutedOpIndices(const MachineInstr &MI,
   return true;
 }
 
+bool MOSInstrInfo::hasCommutePreference(MachineInstr &MI,
+                                        bool &Commute) const {
+  unsigned CommutableOpIdx1 = CommuteAnyOperandIndex;
+  unsigned CommutableOpIdx2 = CommuteAnyOperandIndex;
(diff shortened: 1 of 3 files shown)


llvm/lib/Target/MOS/MOSInstrInfo.cpp
```diff
@@ -267,6 +267,66 @@ bool MOSInstrInfo::findCommutedOpIndices(const MachineInstr &MI,
   return true;
 }
 
+bool MOSInstrInfo::hasCommutePreference(MachineInstr &MI,
```
mysterymath, line 270: This definitely needs some tests that run scenarios through TwoAddressInstruction. It should be fairly straightforward to get it to commute based on the logic here, but if not, then even a simple regression test captured from the original IR would be useful.

mysterymath (commented):
Nice find on the interface for this; another X86-specific gem.

llvm/lib/Target/MOS/MOSInstrInfo.cpp
asiekierka: We already have general tests for TwoAddressInstruction (`commute.mir`). I added a regression test to the set from the relevant issue's IR.

mysterymath (approved):
LGTM!

Issue #454, opened 2024-03-05
[Assembler] Improved ergonomics for 65816 (and other) subtargets


asiekierka:
Note that this issue is about assembler syntax, functionality and ergonomics; it is, generally speaking, *not* about the ELF file format or the compiler backend.

For those who aren't familiar with what LLVM-MOS is - LLVM-MOS is a fork of the LLVM toolchain (which includes a compiler, assembler, linker, and various auxillary tools), capable of efficiently compiling modern C and C++ code to the 6502 and various derivatives with advanced optimization approaches. As such, the focus of "ergonomics" as discussed here is primarily on providing a good developer experience for expressing hand-written code, both for newcomers and experts, across the various relevant enthusiast communities.

## Status quo

The current state of assembler relaxation is explained in great detail [here](https://llvm-mos.org/wiki/Assembler_relaxation) (with the exception of being able to explicitly write out `mos8(...)`, `mos16(...)`, etc); to summarize:

* Constants from 0x0000 to 0x00FF are assumed to be in the zero page,
* Addresses in sections prefixed with special names (like `.zp`) or flags (`z`) are assumed to be in the zero page,
* Symbols marked with `.zeropage <symbol>` are assumed to be in the zero page,
* An explicit cast can be used in the form of `lda mos8(my_address)`.

While the solutions implemented are, in my opinion, sufficient for 6502/65C02 development, further subtargets introduce additional challenges.

Note that I'm going to refer to the zero page/bank page/direct page as "direct page" for all subtargets - that is, the location in memory which addressing modes known on the 6502 as "zero page" refer to:

* The HuC6280 provides its direct page at offset 0x2000, as opposed to 0x0000. We adjusted to this by having the assembler relaxation step consider the 0x2000 - 0x20FF range as the one encoding a one-byte address on this platform.
* The SPC700 allows dynamically moving its direct page to offset 0x0000 or 0x0100. This is presumably meant for stack frame accesses, but can introduce another hurdle in relaxation resolution.
* The 65CE02 allows dynamically moving its direct page to any 256-byte window.
* The 65816... well.

## The beloved 65816

The 65816 introduces *multiple* wrenches to this approach:

* The direct page can be dynamically moved to any location in the bottom 64K of memory.
* 16-bit ("absolute") accesses can point to any 64K of memory. This is decided by the value of the data bank register (for data accesses) or program bank register (for code accesses).
* A new, 24-bit ("long") addressing mode is introduced.
[...]

PR #455, opened 2024-03-09
MOS: 65816 assembler fixes, part 1

asiekierka:
These are the hopefully less controversial parts of https://github.com/llvm-mos/llvm-mos/issues/454 :

* Add support for forcing "long" addressing mode with `lda mos24(addr)`.
* Finish support for `$ml`, `$mh`, `$xl`, `$xh` mapping symbols. These mark the state of `M` and `X` flags required to disassemble a given opcode, allowing tools like `llvm-objdump` to distinguish between `lda #$ea` and `lda #$eaea`. This information is gathered from the instruction itself - `.a8/.a16/.i8/.i16` or whatever we end up implementing to set the default instruction encoding can then simply affect which instruction is selected in the parsing process. Tests have been provided to match other architectures with mapping symbols.
  * I will document these on the wiki as part of the ELF specification (should it be part of the ELF specification?) when this gets merged, don't worry.
* Add missing `BIT #$eaea` opcode to 65816.
* chore: Remove unused/unnecessary imm8To16 and imm16To24 operands. We don't use them anywhere, and it appears that we probably won't.


mysterymath (approved):
Alright, LGTM!

Issue #229, opened 2022-10-07
AXY calling convention


mysterymath:
After getting some experience with NMI code for the NES, a fairly clear calling convention has emerged for such code.

Typically, an interrupt handler will save A, X, and Y, then operate until the handler returns in a context where A, X, and Y are caller-saved, and every other memory resource is callee-saved. This allows arguments to be passed and values to be returned from functions as normal, but without needing to save or restore too much of the interrupted function.

We could adopt this as a formal calling convention for llvm-mos. Code in an interrupt handler could be marked using this calling convention; this would then inform the caller that calls to such functions preserve everything except the registers. Such code could be used efficiently in either interrupt handlers or regular code.

If the function ended up needing some imaginary registers, they would all be considered callee-saved. This would allow the regular whole-program ZP allocation to lift such CSRs to the ZP stack, which would prevent them from needing to be saved and restored. This recovers the typical way of writing interrupt handlers on the 6502: save the regs, and allocate ZP regions dedicated to the handler.

PR #458, opened 2024-03-14
MOS Interrupts: Add the wedge attribute

jtgans:
This allows us to end ISRs with a TailJMP instruction instead of the usual RTI or RTS. This allows for ISRs that are partially defined in the platform's ROM to be continued using a "wedge" style interrupt mechanism, like what is typically required for Commodore machines.

`wedge` takes an argument that is the vector to jump back to, which is the immediate argument to the TailJMP instruction emitted in the call lowering. `wedge` is expected to be used with the normal `interrupt` or `interrupt_norecurse` attributes.

### DIFF
--- a/clang/include/clang/Basic/Attr.td
+++ b/clang/include/clang/Basic/Attr.td
@@ -1869,6 +1869,13 @@ def MOSNoISR : InheritableAttr, TargetSpecificAttr<TargetMOS> {
   let Documentation = [Undocumented];
 }
 
+def MOSWedge : InheritableAttr, TargetSpecificAttr<TargetMOS> {
+  let Spellings = [GCC<"wedge">];
+  let Args = [UnsignedArgument<"ChainISR">];
+  let Subjects = SubjectList<[HasFunctionProto]>;
+  let Documentation = [Undocumented];
+}
+
 def Reentrant : InheritableAttr {
   let Spellings = [GCC<"reentrant">];
   let Subjects = SubjectList<[Function]>;
--- a/clang/lib/CodeGen/Targets/MOS.cpp
+++ b/clang/lib/CodeGen/Targets/MOS.cpp
@@ -54,6 +54,10 @@ public:
       Fn->addFnAttr("interrupt");
     if (FD->getAttr<MOSInterruptNorecurseAttr>())
       Fn->addFnAttr("interrupt-norecurse");
+    if (FD->getAttr<MOSWedgeAttr>()) {
+      auto mosattr = D->getAttr<MOSWedgeAttr>();
+      Fn->addFnAttr("wedge", llvm::utostr(mosattr->getChainISR()));
+    }
     if (FD->getAttr<MOSNoISRAttr>())
       Fn->addFnAttr("no-isr");
   }
--- a/clang/lib/Sema/SemaDeclAttr.cpp
+++ b/clang/lib/Sema/SemaDeclAttr.cpp
@@ -7821,6 +7821,29 @@ static void handleMOSInterruptNorecurseAttr(Sema &S, Decl *D, const ParsedAttr &
   handleSimpleAttribute<MOSInterruptNorecurseAttr>(S, D, AL);
 }
 
+static void handleMOSInterruptWedgeAttr(Sema &S, Decl *D, const ParsedAttr &AL) {
+  if (!isFunctionOrMethod(D)) {
+    S.Diag(D->getLocation(), diag::warn_attribute_wrong_decl_type)
+        << "'wedge'" << ExpectedFunction;
+    return;
+  }
+
+  // Interrupt handler must have void return type.
+  if (!getFunctionOrMethodResultType(D)->isVoidType()) {
+    return;
+  }
+
+  if (!AL.checkExactlyNumArgs(S, 1))
+    return;
+
+  uint32_t ChainISR = 0;
+  if (!checkUInt32Argument(S, AL, AL.getArgAsExpr(0), ChainISR, 0, true))
+      return;
+
+  D->addAttr(::new (S.Context) MOSWedgeAttr(S.Context, AL, (uint16_t)(ChainISR)));
+  D->addAttr(UsedAttr::CreateImplicit(S.Context));
+}
+
 static void handleMOSInterruptNoISRAttr(Sema &S, Decl *D, const ParsedAttr &AL) {
   if (!isFunctionOrMethod(D)) {
     S.Diag(D->getLocation(), diag::warn_attribute_wrong_decl_type)
@@ -9271,6 +9294,9 @@ ProcessDeclAttribute(Sema &S, Scope *scope, Decl *D, const ParsedAttr &AL,
     break;
   case ParsedAttr::AT_BTFDeclTag:
     handleBTFDeclTagAttr(S, D, AL);
+    break;
+  case ParsedAttr::AT_MOSWedge:
+    handleMOSInterruptWedgeAttr(S, D, AL);
     break;
   case ParsedAttr::AT_MOSInterruptNorecurse:
     handleMOSInterruptNorecurseAttr(S, D, AL);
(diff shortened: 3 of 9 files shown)


asiekierka:
Wouldn't this be better solved in the same way we wanted to solve the issue wrt NMI handlers on NES? https://github.com/llvm-mos/llvm-mos/issues/229

That is, adding a new calling convention that makes all registers callee-saved except A, X and Y; in such a situation, your handler would preserve A/X/Y (if necessary), call the appropriately marked function (or functions) in C, then jump to the wedge.

jtgans:
Apologies, I ended up re-editing that reply way more than I probably should have. O.o

asiekierka:
Don't worry about it! I just wanted to understand the problem domain more.

mlund:
> For example, with both the calling convention change _and_ this change, I can do this:

It would be great if you could eventually add a few examples in the SDK, e.g. in `examples/c64` and/or add a section in the wiki.

jtgans:
> > For example, with both the calling convention change _and_ this change, I can do this:
> 
> It would be great if you could eventually add a few examples in the SDK, e.g. in `examples/c64` and/or add a section in the wiki.

Absolutely. I was planning on that when this gets merged in at some point. My platform is typically the C128, but they're similar enough-ish that the C64 is in my crosshairs, too.

In parallel, I'm exploring changing the calling convention to the one described in #229 as well. I'm close, but running into a few issues.

mysterymath:
Hey, sorry to take so long getting around to reviewing this one; a bunch of family stuff came up, and I had to put my llvm-mos maintainer duties on hold for a while.

Looking at this, it introduces some new language-level abstraction, so the natural question as a compiler maintainer is: can the goal already be achieved using a simple combination of existing parts.

Here the following pops into mind:

```
__attribute__((interrupt)) void isr(void) {
  asm volatile ("jmp next_isr");
  __builtin_unreachable();
}
```

This can be used to end any ISR in a JMP rather than a RTI, since the `__builtin_unreachable()` marks whatever follows as dead, and the compiler will then remove the RTI epilogue. This seems broadly to me to have the same effect as `wedge(next_isr)`, but that may just be due to a misunderstanding on my part. (The devil is always in the details, and I haven't spent all that much time looking at this yet.) If I'm mistaken, please let me know, but otherwise, it doesn't seem like this pulls enough weight to be a compiler feature, rather than a piece of advice somewhere on the wiki.

jtgans:
Hah! I didn't even think about using unreachable to achieve the same ends. You're right that the devil here is in the details, which is why I was using TailJMP instead of a straight up JMP to the rest of the ISR.

Does this also omit the postamble for the interrupt? The goal was to enable the compiler to save/restore just the clobbered registers used in the ISR when called, but I worry the unreachable will leave the stack mutated from the preamble save, so that the ROM ISR could function proper.

jtgans:
Personally, I think general code performance is less at issue here -- critical timing getting into and out of the raster interrupt is. At 63 cycles per scanline (on the PAL C64s), you're throwing away a full 19% of a scanline every time the ISR runs with a required JSR and RET. On NTSC it's more, since we get less cycles per scanline and fewer scanlines. At these kinds of slow speeds, even additive effects, er, add up. :P

mysterymath:
> Personally, I think general code performance is less at issue here -- critical timing getting into and out of the raster interrupt is. At 63 cycles per scanline (on the PAL C64s), you're throwing away a full 19% of a scanline every time the ISR runs with a required JSR and RET. On NTSC it's more, since we get less cycles per scanline and fewer scanlines. At these kinds of slow speeds, even additive effects, er, add up. :P

Heh, I suppose everything I said was predicated on the idea that raster interrupts were more-or-less out of scope for a C compiler. When you've got at most 32 instructions to play with, then *any* worse-than-human inoptimality on the part of the compiler becomes a potential bug or feature request.

(We do have an Atari 2600 port, but that seems to me as bit more of a proof of concept or a lark than a serious tool.)

I used to be rosier about llvm-mos getting to "human quality", but that was before I'd spent a significant amt of time looking at the state of the art for, say, ARM. Compilers in general don't provide that degree of optimality. Even without considering super-human superoptimization; every *stage* of compilation is a NP complete that's much easier for a person to solve than a computer, and each uses a heuristic to satisfice. Even breaking it into stages like that isn't optimal. Assembly authors can move concerns in and out of stages freely, violate simplifying assumtions, etc.

As for this change specifically, were it to go in, I do think it would need some more work; it should at the very least allow passing any name that you can give as an compile-time address expression in C (e.g., acceptable as a a global pointer initializer at compile time). The change is already mid-sized; it would probably get larger with that. If it's the only thing needed to make raster interrupts writable in C, that would be one thing, but I'm hesistant to embark on such a journey if I don't think there's actually any way to reach the end goal. (I could be wrong about this though, and maybe our codegen is good enough for most common cases. That would be good to know too.)

jtgans:
> Heh, I suppose everything I said was predicated on the idea that raster interrupts were more-or-less out of scope for a C compiler. When you've got at most 32 instructions to play with, then any worse-than-human inoptimality on the part of the compiler becomes a potential bug or feature request.

I don't understand this mindset. As it stands right now, the llvm-mos wiki mentions that a hand-rolled ISR requires saving all of the registers used if the ISR calls into C. This wastes a full two scanlines on entry and another two on exit saving over 32 registers the compiler uses. Frankly, writing an assembly ISR under these constraints is a bit nuts, and makes the act of using C in the first place quite irritating because to avoid this I have to hand-roll *everything* in assembly that the ISR touches. This essentially eliminates any benefit I get from using llvm-mos in the first place because now nearly half my program is assembly, with duplicated code hand rolled for doing anything more complex than twiddling a register or two.

I should note that both Oscar64 and TRSE have no problems doing raster interrupts in C and Pascal, and I can pretty much guarantee their generated code is worse than llvm-mos.

> As for this change specifically, were it to go in, I do think it would need some more work; it should at the very least allow passing any name that you can give as an compile-time address expression in C (e.g., acceptable as a a global pointer initializer at compile time).

Why, out of curiosity? The address we jump to is not a C function, so storing the pointer in a compile-time address expression seems rather... pointless. C won't be able to make any use of it because it's in the middle of a ROM filled with hand-rolled machine code. Is a `#define` not enough?

> If it's the only thing needed to make raster interrupts writable in C, that would be one thing, but I'm hesistant to embark on such a journey if I don't think there's actually any way to reach the end goal. (I could be wrong about this though, and maybe our codegen is good enough for most common cases. That would be good to know too.)

I can tell you that with this patch, I can do raster ISRs with no trouble on the C128. Without it, interrupts on that platform are totally impossible from C.

mysterymath:
> I don't understand this mindset. As it stands right now, the llvm-mos wiki mentions that a hand-rolled ISR requires saving all of the registers used if the ISR calls into C. This wastes a full two scanlines on entry and another two on exit saving over 32 registers the compiler uses. Frankly, writing an assembly ISR under these constraints is a bit nuts, and makes the act of using C in the first place quite irritating because to avoid this I have to hand-roll _everything_ in assembly that the ISR touches. This essentially eliminates any benefit I get from using llvm-mos in the first place because now nearly half my program is assembly, with duplicated code hand rolled for doing anything more complex than twiddling a register or two.

Sorry, I mis-read what you meant by "raster interrupt". I had thought you meant the type that triggers on horizontal blank; those have no business being written in anything but ASM IMO. For once-per-frame interrupts, then yes, it should be feasible to write them in C.
 
> Why, out of curiosity? The address we jump to is not a C function, so storing the pointer in a compile-time address expression seems rather... pointless. C won't be able to make any use of it because it's in the middle of a ROM filled with hand-rolled machine code. Is a `#define` not enough?

Parsimony and the principle of least surprise. `wedge` accepts an argument with the semantics of an address to jump to after completing the prologue. Without further qualification, any expression that names an address would make sense to put there. Requiring it to be a constant relocatable expression is a ding that creates an exception that needs to be documented; requiring it to be an integer constant is another tighter exception. There also aren't currently any function attributes in llvm-mos that work that way. Less philosophically, requiring an integer constant also precludes symbolic references that can be resolved and managed by the linker (just as usable in assembly as in C).

> I can tell you that with this patch, I can do raster ISRs with no trouble on the C128. Without it, interrupts on that platform are totally impossible from C.

I've been looking for something in this change that tips some project from impossible to possible (since that's sufficient to justify the feature to me), but I still can't see what specifically prevents the wrapping approach from working here. Could you provide a bit more detail on this?

jtgans:
I think we've wasted enough of each other's time on this. Thanks.

Issue #443, opened 2024-02-11
Compilation failure on MacOS w. Apple silicon


mlund:
I get build errors in the last steps around `compiler-rt` on apple M1. Seems `OSX supported arches:` is empty on my setup.

llvm-mos: 97e1882aedf5dba49ae1071c907503036d1c5fc7
cmake: 3.28.1
macos: sonoma 14.3

```console
cmake -C clang/cmake/caches/MOS.cmake -G Ninja -S llvm -B build -DCMAKE_INSTALL_PREFIX=$HOME/llvm-mos
cmake --build build


-- LLVM host triple: arm64-apple-darwin23.3.0
-- LLVM default target triple: mos-unknown-unknown
-- Found libtool - /Users/mikael/github/llvm-mos/build/bin/llvm-libtool-darwin
CMake Deprecation Warning at /Users/mikael/github/llvm-mos/cmake/Modules/CMakePolicy.cmake:6 (cmake_policy):
  The OLD behavior for policy CMP0114 will be removed from a future version
  of CMake.

  The cmake-policies(7) manual explains that the OLD behaviors of all
  policies are deprecated and that a policy should be set to OLD only under
  specific short-term circumstances.  Projects should be ported to the NEW
  behavior and not rely on setting a policy to OLD.
Call Stack (most recent call first):
  /Users/mikael/github/llvm-mos/compiler-rt/CMakeLists.txt:11 (include)


CMake Deprecation Warning at /Users/mikael/github/llvm-mos/cmake/Modules/CMakePolicy.cmake:11 (cmake_policy):
  The OLD behavior for policy CMP0116 will be removed from a future version
  of CMake.

  The cmake-policies(7) manual explains that the OLD behaviors of all
  policies are deprecated and that a policy should be set to OLD only under
  specific short-term circumstances.  Projects should be ported to the NEW
  behavior and not rely on setting a policy to OLD.
Call Stack (most recent call first):
  /Users/mikael/github/llvm-mos/compiler-rt/CMakeLists.txt:11 (include)


-- Found OSX_SYSROOT: /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk
-- Checking DARWIN_iphoneos_SYSROOT - ''
CMake Warning at /Users/mikael/github/llvm-mos/compiler-rt/cmake/base-config-ix.cmake:152 (message):
  iOS SDK not found! Building compiler-rt without iOS support.
Call Stack (most recent call first):
[...]

PR #464, opened 2024-05-16
Use Apple Silicon for CI/CD

### LINKED ISSUE #444: Builder for Apple Silicon
It looks like Github Actions now has beta support for Apple Silicon; we should establish a llvm-mos builder for it.

pr1metine:
Fixes #444. According to [GitHub's list of available GitHub-hosted runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories), we may use M1 machines for our GitHub Actions simply by specifying `runs-on: macos-14`. This PR does just that. In addition to `llvm-project-tests.yml`, `clang-tests.yml`, `lld-test.yaml`, and `llvm-test.yaml` are effected by this as well.

Note that this will most likely result in a broken MacOS pipeline (cf #443). This translates into the inability to build LLVM for Macs.

### DIFF
--- a/.github/workflows/llvm-project-tests.yml
+++ b/.github/workflows/llvm-project-tests.yml
@@ -18,7 +18,7 @@ on:
       python_version:
         required: false
         type: string
-        default: '3.11'
+        default: "3.11"
   workflow_call:
     inputs:
       build_target:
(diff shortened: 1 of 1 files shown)


mysterymath:
I think we need to fix #443 before doing this. We also need to have a strategy for either distributing both x86 and ARM versions of the SDK for Mac, or somehow building LLVM and the SDK using fat binaries. Apple still provides support for x86 Macs, so we should too.

pr1metine:
Hmm yeah I can work on both of these issues

PR #466, opened 2024-06-26
[MOS] BRA, BSR, JSR are not conditional branches

pfusik:


pfusik:
I don't know if that has any effect, but feels right.

mysterymath (approved):
This would be the kind of thing that would pop up as a footgun if we were to directly select one of these  and some later pass choked on it. Thanks!

PR #472, opened 2024-08-08
Legalize signed comparisons with constants to unsigned comparisons

### LINKED ISSUE #470: Faster Signed Comparison
I played a bit with the compiler and found that signed comparisons are generally worse than the unsigned counterpart.
```C
int8_t f(int8_t x)
{
    return x > 32;
}
```
```asm
f(signed char):                                  ; @f(signed char)
        sec
        sbc     #33
        bvc     .LBB0_2
        eor     #128
.LBB0_2:                                ; %entry
        tax
        bpl     .LBB0_4
        lda     #0
        rts
.LBB0_4:                                ; %entry
        lda     #1
        rts
```
But there is an alternative:
```C
int8_t g(int8_t x)
{
    return !(x < 0) && (uint8_t)x > 32;
}
```
```asm
g(signed char):                                  ; @g(signed char)
        tax
        bmi     .LBB1_3
        cmp     #33
        bcc     .LBB1_3
        lda     #1
        rts
.LBB1_3:                                ; %land.rhs
        lda     #0
        rts
```
Both of these sequences do the same, but the second version is shorter and therefore faster (worst case: 6 vs 8 instructions)
https://godbolt.org/z/33z51a8d5

Domkeykong:
took a while until i realized i could use a select and let the lower-select pass introduce the branches.
unfortunately there are some artifacts with the comparisons beyond 8 bits, each byte is compared twice, once for equality and once for Unsigned Greater Equal.

fixes #470


llvm/test/CodeGen/MOS/signed-compare-constant.ll
```diff
@@ -0,0 +1,325 @@
...
+; CHECK-NEXT:  .LBB0_3: ; %entry
+; CHECK-NEXT:    lda #0
+; CHECK-NEXT:    rts
+entry:
+  %cmp = icmp sgt i8 %x, 123
+  %conv2 = zext i1 %cmp to i8
+  ret i8 %conv2
+}
+
+define zeroext i8 @sgt_16(i16 noundef %x) {
```
mysterymath, line 27: Is this still faster than the sequence before this change? If not, it seems better to add a TODO to the legalizer code and restrict the optimization to 8 bits for now. We may need some kind of additional combiner pass to clean up the output of the legalizer, but it's not something where I can guess the details offhand.

llvm/lib/Target/MOS/MOSLegalizerInfo.cpp
```diff
@@ -1206,12 +1207,28 @@ bool MOSLegalizerInfo::legalizeICmp(LegalizerHelper &Helper,
   LLT S1 = LLT::scalar(1);
   LLT S8 = LLT::scalar(8);
 
-  bool RHSIsZero = mi_match(RHS, MRI, m_SpecificICst(0));
+  auto RHSConst = getIConstantVRegValWithLookThrough(RHS, MRI);
+  bool RHSIsZero = RHSConst && RHSConst->Value.isZero();
+
+  if (Pred == CmpInst::ICMP_SLT && RHSConst && !RHSIsZero) {
```
mysterymath, line 1213: A short summary comment about why this is faster than the alternative would be helpful. It's difficult to read out of the implementation.

mysterymath (commented):
Hey, great first change! Have you hacked on LLVM before?

PR #474, opened 2024-09-08
Follow TAX and TAY in combineLdImm

pfusik:


pfusik:
`combineLdImm` knows how to insert TAX and TAY, but doesn't understand them if they are already there.

I can later add the remaining cases (TXA etc).

pfusik:
BTW. why do we have many register-to-register transfer opcodes? Having one would make things simpler here.

mysterymath:
> BTW. why do we have many register-to-register transfer opcodes? Having one would make things simpler here.

This was the most obvious modelling of the transfer instructions such that the register constraints would accurately respect the possibilities. For example, there's no way to express an illegal TXY or TYX instruction, and the MachineVerifier can verify the register constraints for all unit and integration tests. If there were a way to add a target-specific MachineVerifier check for a more general transfer instruction, that might work too, and as you say, it would probably reduce special-casing. I maybe vaguely remember seeing such a thing? Iunno.

llvm/lib/Target/MOS/MOSLateOptimization.cpp
```diff
@@ -261,6 +261,21 @@ bool MOSLateOptimization::combineLdImm(MachineBasicBlock &MBB) const {
   } LoadA, LoadX, LoadY;
 
   for (MachineInstr &MI : MBB) {
+    ImmLoad *Load = nullptr;
+    if (MI.getOpcode() == MOS::TA && LoadA.MI) {
```
mysterymath, line 265: This could use a comment that describes what this code is doing at a higher level; it took me a little bit to relate the code inside with the intent.

(Arguably that's true for the rest of the implementation too, but that's my fault. It was difficult to reacquaint myself with how this code works.)

mysterymath (approved):

pfusik:
> > BTW. why do we have many register-to-register transfer opcodes? Having one would make things simpler here.
> 
> This was the most obvious modelling of the transfer instructions such that the register constraints would accurately respect the possibilities. For example, there's no way to express an illegal TXY or TYX instruction, and the MachineVerifier can verify the register constraints for all unit and integration tests. If there were a way to add a target-specific MachineVerifier check for a more general transfer instruction, that might work too, and as you say, it would probably reduce special-casing. I maybe vaguely remember seeing such a thing? Iunno.

Now I think that having several transfer opcodes is not a big deal here. The assymetry of TXY/TYX explains it.

llvm/lib/Target/MOS/MOSLateOptimization.cpp
pfusik: I can add the comments. I find this routine very straightforward.

pfusik:
Rebased, added comments and removed an unused `Builder` variable.
Please rebase-merge not squash, as the first commit adds a [precommit test](https://llvm.org/docs/InstCombineContributorGuide.html).

pfusik:
> > > BTW. why do we have many register-to-register transfer opcodes? Having one would make things simpler here.
> > 
> > This was the most obvious modelling of the transfer instructions such that the register constraints would accurately respect the possibilities. For example, there's no way to express an illegal TXY or TYX instruction, and the MachineVerifier can verify the register constraints for all unit and integration tests. If there were a way to add a target-specific MachineVerifier check for a more general transfer instruction, that might work too, and as you say, it would probably reduce special-casing. I maybe vaguely remember seeing such a thing? Iunno.
> 
> Now I think that having several transfer opcodes is not a big deal here. The assymetry of TXY/TYX explains it.

`TA` and `T_A` could probably be one opcode. I can try later if it helps.

PR #493, opened 2025-07-19
Relocatable linking does not require a linker file

johnwbyrd:
If the user has given clang the -r option on the command line, then the user is combining multiple object files into one, therefore a linker file MAY not be required, so do not force one into existence.


mysterymath:
Ah, nice catch! LGTM.

PR #501, opened 2025-09-26
Fix incorrect static_casting in MOSAsmParser

johnwbyrd:
The MOSAsmParser constructor receives an MCSubtargetInfo object but unsafely casts it to MOSSubtarget using static_cast. These are fundamentally incompatible types:

- MOSSubtarget inherits from MOSGenSubtargetInfo
- MOSGenMCSubtargetInfo inherits from MCSubtargetInfo
- These have completely different memory layouts and class hierarchies

When MOSAsmParser tried to call methods like hasW65816Or65EL02() on the incorrectly cast object, it accessed member variables (HasW65816Insns, Has65EL02Insns) that were beyond the allocated object boundaries, causing heap buffer overflow.

This problem was discovered with ASAN enabled on the compiler binaries.

### DIFF
--- a/llvm/lib/Target/MOS/AsmParser/MOSAsmParser.cpp
+++ b/llvm/lib/Target/MOS/AsmParser/MOSAsmParser.cpp
@@ -48,13 +48,13 @@ class MOSOperand : public MCParsedAsmOperand {
 public:
   MOSOperand() = delete;
   /// Create an immediate MOSOperand.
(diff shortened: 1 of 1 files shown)


johnwbyrd:
The build problems are crashes in ORC JIT, which is known to have problems on other platforms.

https://github.com/llvm/llvm-project/commit/bd49bbaaafc98433a2cb4e95ce25b7a201baf5a5

Issue #487, opened 2025-05-14
Crash in compiler while writing out relocations


davidgiven:
I've just upgraded llvm-mos, and I'm finding that it's now reliably crashing when building part of CP/M-65. I'm attaching the two files it asked me to attach needed to reproduce this.

[reproduction.zip](https://github.com/user-attachments/files/20206945/reproduction.zip)

It looks like whatever's happening is happening when writing out the output ELF file and has to do with some of llvm-mos' custom relocations. Any ideas? It's making my autobuilder builds fail.

```
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace, preprocessed source, and associated run script.
Stack dump:
0.	Program arguments: /opt/pkg/llvm-mos/bin/clang-21 -cc1as -triple mos-cpm65 -filetype obj -main-file-name osi.S -I /opt/pkg/llvm-mos/bin/../mos-platform/cpm65/asminc -I /opt/pkg/llvm-mos/bin/../mos-platform/common/asminc -I .obj/include/+include_hdr -fdebug-compilation-dir=/home/dtrg/nonshared/cpm65/.obj/src/arch/osi/+osi500mf_bios/src/arch/osi/osi.S/sandbox -dwarf-debug-producer "clang version 21.0.0git (https://github.com/llvm-mos/llvm-mos 83515c2b9e243546c5ecae2e5e643331e7c4481d)" -I /opt/pkg/llvm-mos/bin/../mos-platform/cpm65/asminc -I /opt/pkg/llvm-mos/bin/../mos-platform/common/asminc -I .obj/include/+include_hdr -debug-info-kind=constructor -dwarf-version=5 -mrelocation-model static -object-file-name=/home/dtrg/nonshared/cpm65/.obj/src/arch/osi/+osi500mf_bios/src/arch/osi/osi.S/sandbox/.obj/src/arch/osi/+osi500mf_bios/src/arch/osi/osi.S/osi.o -o .obj/src/arch/osi/+osi500mf_bios/src/arch/osi/osi.S/osi.o /tmp/osi-a1013e.s
 #0 0x00005652bdbb4826 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) (/opt/pkg/llvm-mos/bin/clang-21+0x1366826)
[...]

Issue #457, opened 2024-03-13
[Assembler] .byte/.short don't support MOS expression parsing


asiekierka:
Quoting jroweboy:

> Making split lo/hi LUTs gets very verbose without `<`/`>` or `.lo`/`hibytes` (Keep in mind that you can't split the address using C/C++ code, so ASM is the only way to make an optimized split lo/hi address table, so this is not a contrived example at all. I suspect this would be fairly common in real code)

```asm
; Making a split LUT table
; gnu as gets very verbose for larger tables
LoAddress:
.byte label1@mos16lo, label2@mos16lo
HiAddress:
.byte label1@mos16hi, label2@mos16hi
; ca65
LoAddress:
.byte <label1, <label2  ; < and > can be used as shorthand
.lobytes label1, label2 ; alternatively using lobytes
HiAddress:
.byte >label1, >label2
; alternative defining using define
.define AddressTable label1, label2
LoAddress:
.lobytes AddressTable
HiAddress:
.hibytes AddressTable
```

> Can't use @mos16lo/hi for freestanding values. This came up when generating value tables (i worked around it by just making the generator code cut the hi/lo byte)

```asm
Whatever:
.byte 340@mos16hi ; Does not compile
; ca65 lets you do this though
Whatever:
.byte >340, .hibyte(340)
```

This issue considers only the fact that `.byte <label1, <label2` doesn't work in LLVM-MOS. (Adding support for `.lobytes` or other constructs can be consdered separately.)

One solution to solve this would be to override `.byte`, `.short`, et cetera with our own variants which take MOS expression parsing into account. This is the approach already undertaken by the AVR and MSP430 target backends.

Another solution would be to try and override the expression parser:

> Yeah, it's arguably a wart that our modifiers don't generally apply to expressions (producing relocs in instructions and appropriate directives, and producing constant evaluation effects otherwise). I'd definitely be supportive of work towards that end, assuming that it doesn't require completely refactoring LLVM's parsing/evalutation interfaces for assembler expressions.

While this would solve the problem more generally, there's at least one known  caveat here: WDC syntax (which is what is being implemented as part of https://github.com/llvm-mos/llvm-mos/pull/456 ), as well as the ELF specification, distinguishes between modifiers when used on immediates (so `lda #imm`, but also `.byte`, `.set`) versus used on addresses (`lda addr`, but also `.reloc`).

GorillaSapiens:
i don't get it.  you can't just use

`.short 340`

???

[...]

PR #515, opened 2025-12-10
Issues 457 487

GorillaSapiens:
addresses some issues discussed in #457 and #487

adds the unary <, >, and ^ operators (lo, hi, and bank, respectively) to asm constants
adds the @mos16lo, @mos16hi, and @mos24bank modifiers to asm constants

see changes in 
llvm/test/MC/MOS/modifiers.s b/llvm/test/MC/MOS/modifiers.s
for examples.

NOTE:  see the "TODO FIX" comment in llvm/lib/MC/MCExpr.cpp, i am open to better ways to do this (i'm still learning my way around llvm code here)

### DIFF
--- a/llvm/lib/MC/MCExpr.cpp
+++ b/llvm/lib/MC/MCExpr.cpp
@@ -28,7 +28,8 @@
 #include <cstdint>
 
 // TODO FIX this comes from ./llvm/lib/Target/MOS/MCTargetDesc/MOSMCExpr.h
-// but i cannot figure out how to include it
+// but i cannot figure out how to include it without dragging in a whole
+// bunch of dependencies.
 // this is a GROSS distortion of what it probably should be.
 enum MOSMCExpr {
   VK_NONE,


GorillaSapiens:
i probably need some clues here.  when i 
`cmake --build build --target check-all`
i get
```
CMake Error at /home/gorilla/llvm-mos/fork/llvm-mos/build/lib/cmake/llvm/LLVMExports.cmake:758 (add_library):
  ADD_LIBRARY called with SHARED option but the target platform does not
  support dynamic linking.
Call Stack (most recent call first):
  /home/gorilla/llvm-mos/fork/llvm-mos/build/lib/cmake/llvm/LLVMConfig.cmake:261 (include)
  CMakeLists.txt:63 (find_package)
```

i suspect it's how i configure and/or build.  is this correct?
```
cmake -C clang/cmake/caches/MOS.cmake -G Ninja -S llvm -B build
cmake --build build --target distribution

```

GorillaSapiens:
ok, gleaned the correct commands from the log.  looks like i broke a ppc test and an x86_64 test.  and yes, i know i touched non-MOS code.  do we care?  or should i go back to the drawing board and figure out how to do this w/o breaking that?

llvm/lib/MC/MCExpr.cpp
```diff
@@ -216,6 +242,47 @@ const MCConstantExpr *MCConstantExpr::create(int64_t Value, MCContext &Ctx,
   return new (Ctx) MCConstantExpr(Value, PrintInHex, SizeInBytes);
 }
 
+const MCConstantExpr *MCConstantExpr::create(int64_t Value,
```
mysterymath: It isn't appropriate for any MOS code to live in MCExpr.cpp; that's expected to be completely target indepentent. You'll need to a way to implement this that doesn't place any MOS-namespace code here. It also seems rather strange to have MCConstantExpr go from being an int64_t wrapper to something like... this.

PR #516, opened 2025-12-16
Add assembler support for <, >, and ^ operators

GorillaSapiens:
related to issues #457 and #487

for MOS only, add assembler support for <, >, and ^ operators on constants, as well as @mos16lo, @mos16hi, and @mos24bank modifiers on constants, in both instructions and .byte statements.

### DIFF
--- a/.github/workflows/llvm-project-tests.yml
+++ b/.github/workflows/llvm-project-tests.yml
@@ -14,7 +14,7 @@ on:
         required: false
       os_list:
         required: false
-        default: '["ubuntu-24.04", "windows-2022", "macOS-latest"]'
+        default: '["ubuntu-24.04", "windows-2022", "macOS-15"]'
       python_version:
         required: false
         type: string
@@ -39,7 +39,7 @@ on:
         type: string
         # Use windows-2019 due to:
         # https://developercommunity.visualstudio.com/t/Prev-Issue---with-__assume-isnan-/1597317
-        default: '["ubuntu-24.04", "windows-2022", "macOS-latest"]'
+        default: '["ubuntu-24.04", "windows-2022", "macOS-15"]'
 
       python_version:
         required: false


johnwbyrd:
I can't tell if this code is 100% correct, but I do think it's an improvement, so I'm going to let it roll.  Thanks for the submission.

mysterymath (commented):
Sorry for sleeping on this one; I had unfortunately confused it for the prior change. A cursory looks through this looks real good; thanks for putting this together. Very clever, yet clean way to go about it. Would you mind updating the wiki for the assembler; I think it's out of date now. I'll cut a release soon for this; it's been a while.

GorillaSapiens:
> Sorry for sleeping on this one; I had unfortunately confused it for the prior change. A cursory looks through this looks real good; thanks for putting this together. Very clever, yet clean way to go about it. Would you mind updating the wiki for the assembler; I think it's out of date now. I'll cut a release soon for this; it's been a while.

@mysterymath 

are we talking about editing the "Target-specific-directives" table in https://llvm-mos.org/wiki/Assembler ?
i've created an account, but i'm not seeing any obvious way to edit pages.
perhaps i need some permissions to do that ?
or should i be looking somewhere else entirely?

mysterymath:
@johnwbyrd ; is there some requirement for who can edit?

mysterymath:
> are we talking about editing the "Target-specific-directives" table in https://llvm-mos.org/wiki/Assembler ? i've created an account, but i'm not seeing any obvious way to edit pages. perhaps i need some permissions to do that ? or should i be looking somewhere else entirely?

I was thinking of the page below. For example: "The above syntax for modifiers only works inside assembly language instructions. If you need to use modifiers in assembly directives, the syntax is instead: .byte address@mos16hi (or similar)."

https://llvm-mos.org/wiki/Modifiers

GorillaSapiens:
@mysterymath i'm also not sure whether this addresses ALL the issues in #457 and #487, or just some of them.

GorillaSapiens:
> @johnwbyrd ; is there some requirement for who can edit?

LOL, i found it, i needed to "verify" my email address.

PR #519, opened 2025-12-26
feat(MOS): Add source-level debugging support

johnwbyrd:
## Summary

This PR adds source-level debugging capabilities to llvm-mos, enabling developers to debug MOS programs using LLDB with GDB stub debuggers (e.g., MAME).

Key components:
- DWARF register numbering and CFI (Call Frame Information) generation for MOS
- LLDB ABI plugin for MOS with custom register context handling
- Garbage collection support for `.debug_frame` sections in LLD
- Infrastructure improvements to GlobalISel, MIR parsing, and AsmPrinter

### MOS-Specific
- Add DWARF register numbers per MOS specification (A, X, Y, S, P, PC, imaginary registers)
- Implement CFI emission in MOSFrameLowering for dual-stack architecture
- Add `getFrameIndexReference` for debug info frame base resolution
- Expand floating-point libcall support and custom lower `modf`
- Fix jump table size limit to 255 entries
- Disable `--gc-sections` for relocatable links

### LLDB
- Add ABI extension points for custom register contexts (`CreateRegisterContextForThread`, `ProvidesRegisterInfoOverride`)
- Implement MOS ABI plugin with register definitions and unwind support
- Add lldb, lldb-dap, llvm-dis, llvm-config to MOS distribution

### LLD
- Support garbage collection of `.debug_frame` sections
- Handle `R_MOS_NONE` relocations in MOS::getRelExpr

### Infrastructure (upstream candidates)
- Allow deferring AsmPrinter handler initialization via `shouldCallBeginModule()`
- Implement debug info salvaging for `G_MERGE_VALUES` in GlobalISel
- Support parsing/printing complex CFI instructions in MIR (cfi_val_offset, cfi_def_cfa_expression, etc.)
- Improve diagnostics for inline asm constraint errors

### Testing
- Add semantic lit features (`global-isel-only`, `no-selectiondag-debug`, `aggressive-tailcall`, etc.)
- Mark generic tests unsupported on MOS where appropriate

## Test plan

- [ ] `cmake --build build --target check-llvm` passes
- [ ] `cmake --build build --target check-lld` passes
- [ ] Manual debugging session with MAME GDB stub works


mysterymath:
> Yes, well, that is the point of each of the commits in the series. Each commit is intended to be atomic, to the extent that I have been able to make it so. If you would prefer each commit to be broken into a separate pull, that is possible, but many of the later commits are dependent such that the earlier ones need to land before the later ones. Do you want this to become interdependent PRs?

Yeah, it really needs to be PRs that are as independent as possible; otherwise it's quite difficult to land parts individually, and more importantly, it's quite difficult to revert parts individually if we discover problems with it. Separate PRs also gives separate forums for discussion, and many many of these aspects can, and should be, completely independently discussed.

This is also the policy of upstream LLVM, for the same reasons (https://llvm.org/docs/DeveloperPolicy.html#incremental-development). They actually go further, and they discourage working on separate branches at all as you and I have been doing. But it's understandable for things as researchy as this; development would create a lot of upstream churn in such cases.

Stacking PRs on one another is annoying to do (but possible), so I'd be hesistant to make it policy. Given that we have this uber change, it's just as straightforward to hold off on uploading dependent PRs until their dependency PRs have been landed. If we keep rebasing *this* change on head, it will go smaller and smaller until eventually landed. This can then act as a tracker for the overall work, and still allow people to try the feature while it's being polished and stably landed.

Looking at the commits above, it's definitely *not* the case that every commit in the chain depends on all the ones immediately previous. Actually working out what the dependency chains are and encoding that in PR sets would help us get this safely in as fast as possible. (As we can be working on landing many independent changes at the same time, with minimum bottlenecking.)

johnwbyrd:
Yes, that's an accurate description of the state of things.  Some of the commits will be controversial, which I expect and am ready to discuss; some of them are just plain old bug fixes; but I believe that I ran head-first into all these problems while bringing up source-level debugging.  I'll close this PR and resubmit as individual PRs, starting with the PRs that shouldn't be terribly hard.

And, likewise, you need never tl;dr: me.  Everything you write is insightful and relevant.

mysterymath:
Oops, the tldr was just because GitHub wouldn't let me hit the red button without a comment. No tone intended; should have just used "."

PR #520, opened 2025-12-30
fix(transforms): handle null return from emitMemCpyChk

johnwbyrd:
## Summary
- Handle null return from `emitMemCpyChk` in `optimizeStrpCpyChk` before using the result
- Previously the code would call `cast<CallInst>(nullptr)` causing an assertion failure
- This is exposed by targets like MOS that lack `__memcpy_chk` libcall


johnwbyrd:
I'm not going to do that for the size of the fix and the scope of the code changed.  The code clearly doesn't handle the error case correctly.

mysterymath (approved):
Oh, I wasn't reading this right when I skimmed it for my first pass. The original code is *clearly* bogus. As an ambassador for LLVM, consider sending this PR upstream (presuming that the blood has not yet become too bad).

PR #521, opened 2025-12-30
fix(lldb): handle empty data in DataExtractor::Checksum

johnwbyrd:
## Summary
- Fix crash when Checksum() is called with empty or null data buffer
- LLDB was crashing when connecting to MOS target due to empty data being passed to MD5
- Returns early with cleared dest if no data to checksum

### DIFF
--- a/lldb/source/Utility/DataExtractor.cpp
+++ b/lldb/source/Utility/DataExtractor.cpp
@@ -1030,6 +1030,12 @@ void DataExtractor::Checksum(llvm::SmallVectorImpl<uint8_t> &dest,
   else
     max_data = std::min(max_data, GetByteSize());
 
+  // Give up early if no data to checksum
+  if (max_data == 0 || GetDataStart() == nullptr) {
+    dest.clear();
+    return;
+  }
+
   llvm::MD5 md5;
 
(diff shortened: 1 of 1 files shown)


johnwbyrd:
I believe that this crash will affect any target attempting to connect to a gdb remote with incomplete register information.  Fixing this crash was not gating to get lldb working for MOS -- I did in fact work around it in later commits.  Getting a call stack was a pain in the ass though, because you have to debug lldb with lldb in order to find it.  If this fix offends you, we can defer it and/or change it to an assert.

mysterymath:
> I believe that this crash will affect any target attempting to connect to a gdb remote with incomplete register information. Fixing this crash was not gating to get lldb working for MOS -- I did in fact work around it in later commits. Getting a call stack was a pain in the ass though, because you have to debug lldb with lldb in order to find it. If this fix offends you, we can defer it and/or change it to an assert.

That's fair; I think this one is probably fine on second thought. It should always be safe to define an undefined behavior, and if the signal we were getting wasn't useful, then moving the breakage somewhere higher in the stack seems fine.

mysterymath (approved):

PR #524, opened 2025-12-30
fix(lldb): correct MOS address size to 2 bytes

johnwbyrd:
## Summary
- MOS is a 16-bit address space architecture
- The address size was incorrectly set to 4 bytes, now corrected to 2 bytes


johnwbyrd:
As with every change in this entire parade of changes, they are necessary to get DWARF working.  DWARF parsing, memory operations, stack unwinding, and symbol resolution all depend upon this value.  Support for those other platforms is entirely theoretical at this moment.

PR #525, opened 2025-12-30
fix(MOS): don't pass --gc-sections for relocatable links

johnwbyrd:
## Summary

Don't pass `--gc-sections` to the linker when using `-r` (relocatable link) to combine object files. Garbage collection is inappropriate for partial links since sections that appear unused may be needed when the combined object is later linked into the final binary.

### DIFF
--- a/clang/lib/Driver/ToolChains/MOSToolchain.cpp
+++ b/clang/lib/Driver/ToolChains/MOSToolchain.cpp
@@ -69,7 +69,10 @@ void mos::Linker::ConstructJob(Compilation &C, const JobAction &JA,
 
   // Pass defaults before AddLinkerInputs, since that includes -Wl
   // options, which should override these.
-  CmdArgs.push_back("--gc-sections");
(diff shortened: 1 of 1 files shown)


johnwbyrd:
Again, it's clear from the documentation that --gc-sections and -r are incompatible, yet --gc-sections is always imposed upon the user.  This is in error.

mysterymath:
One finer point I wanted to look up here is that sometimes you actually *do* want to use `--gc-sections` with `-r`; there's a use for that. But it's a niche, and more imporantly, it requires some additional work to make meaningful. So we absolutely mustn't spring a `--gc-sections` on a user that specified `-r`; that's ruinously surprising. This should go in asap.

I also think that this is small enough to go in without tests; we don't *really* expend much effort testing the compiler driver; as it's pretty simple. We may need to change that as we continue to make it more complex though.

mysterymath (approved):

Issue #517, opened 2025-12-18
Exactly 256 switch cases crashes compiler


johnwbyrd:
A switch with precisely 256 cases will crash the compiler.

mos-clang -c /tmp/test_jumptable_256.c -o /tmp/test_jumptable_256.o

MOSISelLowering.cpp sets setMaximumJumpTableSize(256) but the backend's jump table implementation requires < 256 entries due to the split lo/hi byte table format with 8-bit indexing.

[test_jumptable_256.c](https://github.com/user-attachments/files/24228050/test_jumptable_256.c)

PR #528, opened 2025-12-30
fix(MOS): limit jump tables to 255 entries and fix getNumRegisters

johnwbyrd:
## Summary

Two fixes for MOS instruction selection:

1. **Jump table limit**: Reduce maximum jump table size from 256 to 255 entries. MOS jump tables use split low/high byte arrays indexed by an 8-bit register. With 256 entries, the offset to access the high byte array would overflow 8-bit arithmetic.

2. **getNumRegisters**: Use ceiling division `(bits + 7) / 8` instead of floor division `bits / 8`. This ensures non-power-of-2 types like i9 get enough registers (2 instead of 1).

### DIFF
--- a/llvm/lib/Target/MOS/MOSISelLowering.cpp
+++ b/llvm/lib/Target/MOS/MOSISelLowering.cpp
@@ -48,9 +48,8 @@ MOSTargetLowering::MOSTargetLowering(const MOSTargetMachine &TM,
   setStackPointerRegisterToSaveRestore(MOS::RS0);
 
   // MOS jump tables use split low/high byte arrays indexed by an 8-bit register.
-  // To access the high byte array, the index is offset by the table size.
-  // With 256 entries, this offset would overflow 8-bit arithmetic, so limit to 255.
-  setMaximumJumpTableSize(std::min(255u, getMaximumJumpTableSize()));
+  // The 8-bit index register can hold values 0-255, allowing up to 256 entries.
+  setMaximumJumpTableSize(std::min(256u, getMaximumJumpTableSize()));
 }
 
 bool MOSTargetLowering::isSuitableForJumpTable(const SwitchInst *SI,
@@ -61,7 +60,7 @@ bool MOSTargetLowering::isSuitableForJumpTable(const SwitchInst *SI,
   // MOS jump tables use split low/high byte arrays indexed by an 8-bit register.
   // This is a hard architectural limit that must be enforced regardless of
   // optimization level (the base class bypasses MaxJumpTableSize for -Os).
-  if (Range > 255)
+  if (Range > 256)
     return false;
   return TargetLowering::isSuitableForJumpTable(SI, NumCases, Range, PSI, BFI);
 }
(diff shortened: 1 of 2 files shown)


mysterymath:
> > ```c++
> >     assert(Table.MBBs.size() < 256);
> >     Expr = MCBinaryExpr::createAdd(
> >         Expr, MCConstantExpr::create(Table.MBBs.size(), Ctx), Ctx);
> > ```
> 
> i agree that assert looks suspect also. shouldn't it be <= 256 ?

Yep, 100% this is assert is wrong; should be <= 256. My booboo.

johnwbyrd:
Nevertheless y'all, codegen currently dies when it tries to deal with a switch with 256 cases, or with BitInt with sizes that cannot be divided by 8.

llvm/lib/Target/MOS/MOSISelLowering.cpp
```diff
@@ -47,7 +47,23 @@ MOSTargetLowering::MOSTargetLowering(const MOSTargetMachine &TM,
...
+
+bool MOSTargetLowering::isSuitableForJumpTable(const SwitchInst *SI,
+                                               uint64_t NumCases,
+                                               uint64_t Range,
+                                               ProfileSummaryInfo *PSI,
+                                               BlockFrequencyInfo *BFI) const {
+  // MOS jump tables use split low/high byte arrays indexed by an 8-bit register.
+  // This is a hard architectural limit that must be enforced regardless of
+  // optimization level (the base class bypasses MaxJumpTableSize for -Os).
+  if (Range > 255)
```
mysterymath: Can we reuse the maximum jump table size from the above rather than 255? This would give us one place to set the constant.

johnwbyrd:
Fixes https://github.com/llvm-mos/llvm-mos/issues/517 .

johnwbyrd:
I've pushed an additional commit that increases the jump table limit from 255 to 256 entries.

### Analysis

The original 255 limit was based on a concern about 8-bit arithmetic overflow when computing the high byte array offset. However, this concern was misplaced:

- The offset to access the high byte array (`table + N`) is computed at assembly/link time as a constant address, not as runtime 8-bit arithmetic
- The 8-bit index register (X or Y) can hold values 0-255, which addresses exactly 256 entries

### Changes in the new commit

1. `setMaximumJumpTableSize(std::min(256u, ...))` (was 255)
2. `if (Range > 256)` in `isSuitableForJumpTable` (was 255)
3. `assert(Table.MBBs.size() <= 256)` (was `< 256`)
4. Updated comments to remove the incorrect overflow explanation

### Testing

| Test Case | Range | Result |
|-----------|-------|--------|
| 255 cases | 255 | Single 256-entry jump table |
| 256 cases | 256 | Single 256-entry jump table |
| 257 cases | 257 | 256-entry JT + case 0 handled separately |
| 258 cases | 258 | 256-entry JT + extra cases handled separately |

The compiler correctly splits larger switches, using up to 256 entries per jump table.

mysterymath (approved):

PR #530, opened 2025-12-30
feat(MOS): expand floating-point libcall support

johnwbyrd:
## Summary

Expand floating-point libcall support in the MOS legalizer:

- Add more trig functions: `tan`, `acos`, `asin`, `atan`, `atan2`, `cosh`, `sinh`, `tanh`
- Add `fexp10`, `flog10`
- Add `fminimumnum`, `fmaximumnum` (IEEE 754-2019 minimum/maximum)
- Add `fsincos` for combined sin/cos
- Add custom lowering for `modf` intrinsic (requires output pointer handling)
- Clean up duplicate `G_FMINNUM` entry and unused includes

### DIFF
--- a/llvm/test/CodeGen/MOS/legalizer.mir
+++ b/llvm/test/CodeGen/MOS/legalizer.mir
@@ -3250,3 +3250,125 @@ body: |
     %3:_(s8) = G_TRUNC %2
     RTS implicit %3
 ...
+---
+name: modf_f32
+body: |
+  bb.0.entry:
+    ; CHECK-LABEL: name: modf_f32
+    ; CHECK: [[COPY:%[0-9]+]]:_(p0) = COPY $rs0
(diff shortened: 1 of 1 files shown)


llvm/lib/Target/MOS/MOSLegalizerInfo.cpp
```diff
@@ -409,6 +421,46 @@ bool MOSLegalizerInfo::legalizeIntrinsic(LegalizerHelper &Helper,
     MI.eraseFromParent();
     return true;
   }
+  case Intrinsic::modf: {
```
mysterymath: This is definitely complex enough to have its own tests in legalizer.mir.

johnwbyrd: Okay, I think I (just) added them to this PR.

PR #533, opened 2026-01-01
More modifier tests

GorillaSapiens:
allow function style modifiers in assembly directives.  these are valid now:

```
.byte $a9, mos16lo(1234)        ; CHECK: a9 d2
.byte $a9, mos16lo($1234)       ; CHECK: a9 34

.byte $a9, mos16hi(1234)        ; CHECK: a9 04
.byte $a9, mos16hi($1234)       ; CHECK: a9 12

.byte $a9, mos24bank(1234)      ; CHECK: a9 00
.byte $a9, mos24bank($1234)     ; CHECK: a9 00
```

### DIFF
--- a/llvm/lib/Target/MOS/MCTargetDesc/MOSModifierNames.cpp
+++ b/llvm/lib/Target/MOS/MCTargetDesc/MOSModifierNames.cpp
@@ -1,3 +1,11 @@
+//===-- MOSModifierNames.cpp - MOS modifier name table --------------------===//
+//
+// Part of LLVM-MOS, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
 #include "MOSModifierNames.h"
 #include "llvm/ADT/ArrayRef.h"
 
--- a/llvm/lib/Target/MOS/MOSModifierNames.h
+++ b/llvm/lib/Target/MOS/MOSModifierNames.h
@@ -1,4 +1,13 @@
-#pragma once
+//===-- MOSModifierNames.h - MOS modifier name table ------------*- C++ -*-===//
+//
+// Part of LLVM-MOS, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef LLVM_LIB_TARGET_MOS_MOSMODIFIERNAMES_H
+#define LLVM_LIB_TARGET_MOS_MOSMODIFIERNAMES_H
 
 #include "MCTargetDesc/MOSMCExpr.h"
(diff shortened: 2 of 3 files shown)


GorillaSapiens:
> I see what you are doing here, and I am basically on board with the refactoring. Give me a bit to look at this.

yes, it seemed poor form to just copy and paste the same array into another file.  so i had to put it where both files could get at it.

llvm/lib/Target/MOS/AsmParser/MOSAsmParser.cpp
```diff
@@ -322,6 +323,47 @@ class MOSAsmParser : public MCTargetAsmParser {
...
+    // Accept mos16lo(expr), mos16hi(expr), mos24bank(expr) in directives.
+    // The generic MC expression parser does not support function-call syntax,
+    // so we translate these into MOSMCExpr variants here.
+    if (Tok.getKind() == AsmToken::Identifier) {
+      StringRef Name = Tok.getString();
+
+      MOSMCExpr::VariantKind VK = MOSMCExpr::VK_NONE;
+
+      const auto &Modifier =
+          std::find_if(std::begin(MOS::modifierNames()), std::end(MOS::modifierNames()),
```
mysterymath: `llvm::find_if` in `ADT/STLExtras.h` accepts ranges directly, which is easier to read.

llvm/lib/Target/MOS/MCTargetDesc/MOSModifierNames.cpp
```diff
@@ -0,0 +1,28 @@
+#include "MOSModifierNames.h"
+#include "llvm/ADT/ArrayRef.h"
+
+namespace llvm {
+namespace MOS {
+
+  static const struct ModifierEntry ModifierNames[] = {
+    // Define immediate variants of mos8() and mos16() first.
```
mysterymath: In LLVM style, these should be indented even with the namespace.

Here and throughout.

llvm/lib/Target/MOS/MCTargetDesc/MOSMCExpr.cpp
```diff
@@ -189,10 +166,10 @@ void MOSMCExpr::visitUsedExpr(MCStreamer &Streamer) const {
 
 const char *MOSMCExpr::getName() const {
   const auto &Modifier = std::find_if(
```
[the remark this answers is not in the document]
GorillaSapiens: to be fair, some of this is not code i touched, it was already using std::find_if.

johnwbyrd:
Sorry, I'll get on this.

johnwbyrd:
Okay, a few things, which are pretty minor, but I would ask you to note, so future pulls from you follow these patterns.

First: LLVM doesn't #pragma once -- there's a standard header and include guard we use for new files -- please use that.  I fixed this.

Second: you didn't round out the test suite with tests covering every use case of function-style modifiers.  I tried to add everything, but you should really think through every use case and get them all into the build.

You will probably consider these changes to be nits.  Unfortunately, all these minor stylistic issues is the price we pay working in the LLVM domain.  @mysterymath and I have been fairly persnickety in terms of what we've allowed into the fork, and as a result everything kinda sorta works as it should.  I am asking you politely to play along.

PR #537, opened 2026-01-04
test: add MOS lit test features and mark unsupported tests

johnwbyrd:
## Summary

Add MOS-specific lit feature flags and mark tests that use unsupported features as `UNSUPPORTED` on MOS.

## Background

The MOS target has several architectural limitations compared to typical LLVM targets:

- **GlobalISel-only**: MOS uses GlobalISel exclusively without SelectionDAG fallback
- **No exception handling**: No invoke/landingpad support
- **No vector types**: Limited register file cannot support vectors
- **Limited inline asm**: Restricted inline assembly support due to few registers
- **16-bit address space**: Pointers are 16-bit, not 32/64-bit

Many generic LLVM tests assume features that MOS doesn't support, causing spurious test failures.

### 2. Mark unsupported generic tests

Added `UNSUPPORTED:` markers to tests that use features MOS doesn't support:

| Feature | Tests Marked |
|---------|-------------|
| `global-isel-only` | 8 tests using SelectionDAG-specific behavior |
| `no-invoke-support` | 2 tests using invoke/exception handling |
| `no-vector-legalization` | 4 tests using vector types |
| `limited-inline-asm` | 4 tests using complex inline asm |
| `no-selectiondag-debug` | 6 tests using `-debug` with SelectionDAG |
| `small-address-space` | 1 test assuming 32-bit addresses |

### 3. Fix DWARF test utilities (`DwarfUtils.cpp`)

The DWARF unit test utilities (`getDefaultTargetTripleForAddrSize`) assert that address size must be 32 or 64 bits. This fails when the default target is a 16-bit architecture.

Added early return for 16-bit architectures (AVR, MOS, MSP430) to return `UnknownArch`, allowing these tests to skip gracefully.

## Result

Clean test runs on MOS target without spurious failures from tests that exercise unsupported features.


johnwbyrd:
Sure.  It seems to me that check-llvm is broken when run on llvm-mos.  There are many tests which cannot possibly run on GlobalISel only systems.  I am trying to set up categories of tests so that I can, in an upcoming checkin, mark tests as unsupported on not only llvm-mos, but any platform that uses GlobalISel without a SelectionDAG fallback. With this, and the upcoming DWARF deltas, check-llvm passes for llvm-mos.

johnwbyrd:
Thinking about this more, I've come around to your point. The tests I was marking UNSUPPORTED are testing target-independent LLVM machinery -- GlobalISel infrastructure, invoke handling, vector legalization, generic inline asm, small-address-space. Running them on X86/AArch64 gives identical coverage; my marking exercise was just making `check-llvm` exit 0 without adding real regression protection. And the ongoing maintenance -- catching each new upstream test that trips the same categories -- is a burden neither of us should be signing up for.

Your alternative is cleaner: encode "build MOS but run check-llvm against a host target" into MOS.cmake so contributors get the correct signal by default, with MOS-specific behavior tests continuing to live in `llvm/test/CodeGen/MOS/` where they get exercised naturally.

Since you already maintain the alternative config that makes the tests pass, would you be willing to encode it into MOS.cmake? I don't know what settings you're using, so I'd be guessing if I tried to sketch it myself.

Once that lands, I'll close #537 as superseded.

Issue #128, opened 2021-12-09
Push/pull elision


mysterymath:
Register scavenging can produce chains of PHA PLA PHA PLA that could be coalesced into a single PHA PLA.

Alternatively, if we can determine that the original value of A was an immediate load, then we can copy rematerialize the load at the site of the last PLA. If the PHA kills A, then the PHA can also be deleted.

mysterymath:
Closing as obsolete; this doesn't appear to come up very often at all.

PR #552, opened 2026-02-18
[65CE02] Use ASR instruction for arithmetic shift right

mlund:
## Summary

- Selects the 65CE02 `ASR` instruction for signed right shifts (`ashr`), replacing the `CMP #128` + `ROR` idiom
- Skips ASR when the result is dead (store-folded RMW), avoiding register pressure regression
- Gated on `has65CE02()` — no effect on other targets

## C/C++ benefit

Any `int8_t >> n`, `int16_t >> n`, or `int32_t >> n` where `n` is a small constant. The high byte of every signed right shift uses ASR instead of two instructions. Common in:

- Audio mixing (sample attenuation: `sample >>= 1`)
- Fixed-point arithmetic (`fixed_t result = (a * b) >> 8`)
- Sprite/tile coordinate scaling
- Signed division by powers of two

## Codegen improvement (vs. baseline, zero regressions)

```
Function         | Instructions | Bytes
-----------------+--------------+------
ashr_i8_1        |      -1      |  -2
ashr_i8_2        |      -2      |  -4
ashr_i16_1       |      -3      |  -4
ashr_i32_1       |      -5      | -10
ashr_global_i16  |       0      |   0
all others       |       0      |   0
```

All 117 MOS tests pass (78 CodeGen, 39 MC).

## Test plan

- [x] `llvm/test/CodeGen/MOS/asr-65ce02.ll` — positive and negative cases
- [x] Full MOS CodeGen + MC test suite passes
- [x] Verified no regression on store-folded globals

Co-authored with Claude Opus 4.6 and humanly supervised and reviewed
by @mlund in accordance with LLVM guidelines.

### DIFF
--- a/llvm/lib/Target/MOS/MOSInstructionSelector.cpp
+++ b/llvm/lib/Target/MOS/MOSInstructionSelector.cpp
@@ -1687,6 +1687,8 @@ bool MOSInstructionSelector::selectMergeValues(MachineInstr &MI) {
 
 bool MOSInstructionSelector::selectLshrShlE(MachineInstr &MI) {
   auto [Dst, CarryOut, Src, CarryIn] = MI.getFirst4Regs();
+  MachineIRBuilder Builder(MI);
+  const auto &MRI = *Builder.getMRI();
 
   unsigned ShiftOpcode, RotateOpcode;
   switch (MI.getOpcode()) {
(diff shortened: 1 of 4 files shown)


llvm/lib/Target/MOS/MOSInstructionSelector.cpp
```diff
@@ -1697,13 +1699,35 @@ bool MOSInstructionSelector::selectLshrShlE(MachineInstr &MI) {
     RotateOpcode = MOS::ROL;
     break;
   case MOS::G_LSHRE:
+    // 65CE02 ASR replaces the CMP #128 + ROR idiom for arithmetic right
```
mysterymath: I'd give this to Claude: this *might* be more elegantly solved in the legalizer instead. The line between legalization and instruction selection is... definitely fuzzy, but ASR feels more like an operation that is directly legal on the 65CE02 than a pattern of other, legal operations.

Actually yeah, if you actually wanted to *select* this pattern, it would be: shift right, but the carry in is the sign bit of the thing you're rotating. This only detects one way where that might be true... the general case should an abstract logic over values! Much more legalizer-ey.

mlund: I'll try to transfer this to legalizer, closing.

PR #553, opened 2026-03-13
[MOS] Add TRI hook to allow remat past register class restrictions

koalabuttz:
Full disclosure, I heavily used Claude Opus 4.6 to assist diagnosing this and writing the code. I reviewed the LLVM project's LLM Policy and have made a genuine effort at understanding the change.

My understanding is that `rematWillIncreaseRestriction` was written with AMDGPU in mind, a target where register class widening from live range splitting is beneficial. This function prevents rematerialization from undoing that. However, the check runs for all targets, even when widening doesn't serve the same purpose.

On MOS, `getLargestLegalSuperClass` returns `Anyi8`, which is a union of `GPR` (the hardware A/X/Y registers) and `Imag8` (zero-page memory locations). As I understand it, these two sets don't share any physical registers, so "widening" to `Anyi8` isn't actually useful to the allocator. But because rematerialized instructions like `LDImm` constrain to `GPR` (a subclass of `Anyi8`) the check treats every rematerialization as a restriction and blocks it. The result is that cheap immediate reloads get replaced with expensive multi-byte COPY/spill sequences.

**Real-world impact:** On a C64 roguelike with ~37 KB of code in ~50 KB of usable RAM, the blocked remats added 1,591 bytes to `.text` and dropped free RAM from 2,523 to 930 bytes. The effect scales with binary size — the isolated test IR only showed 468 bytes.

This fix adds a virtual method `shouldBlockClassRestrictingRemat()` to `TargetRegisterInfo` that defaults to `true`, preserving existing behavior for all targets. MOS overrides it to return `false` to opt out of the check. A MIR test verifies that `LDImm` instructions are rematerialized after a call rather than spilled.

I went with a simple target hook rather than something more automatic because it seemed like the safest approach and I didn't want to risk changing behavior for backends I don't understand well enough.

Please share any feedback you have on this PR — I would like for this and any future contributions of mine to be productive and align with the project goals.


mysterymath:
Hah, interesting; I just landed a fix WRT this code, as it was causing a crash: https://github.com/llvm-mos/llvm-mos/pull/558

I don't think it's generally the case that widening from GPR to Anyi8 is useless; far from it. GPR is much tighter to manage than Anyi8 is. Accordingly, the justification for this change doesn't seem accurate; I'm not convinced that we differ materially from AMDGPU. This routine is somewhat newer though: we should investigate its original intent, and perhaps modify it to be less conservative.

PR #567, opened 2026-07-06
feat(lld/ELF): support garbage collection of .debug_frame sections

johnwbyrd:
## Summary

Add `DebugFrameInputSection` and `DebugFrameSection` to handle `.debug_frame`
similarly to `.eh_frame`. When `--gc-sections` is used, FDEs whose functions
were garbage-collected are discarded rather than left with dangling
relocations.

## Background

MOS uses `--gc-sections` by default. When compiling with CFI enabled (`-g` or
explicit CFI), each function emits an FDE in `.debug_frame` that references
the function's start via a relocation. If GC then discards the function's
`.text` section, LLD emits relocation errors against the deleted symbol.

`.eh_frame` already has this handling. `.debug_frame` is structurally similar
but not identical:

- CIE is identified by `0xFFFFFFFF` (vs. `0` for `.eh_frame`)
- CIE pointer in FDE is an absolute offset from section start (vs. relative
  for `.eh_frame`)
- Not loaded at runtime — `SHT_PROGBITS`, no `SHF_ALLOC`
- FDE's function reference is the second relocation (offset 8), not the
  first (offset 0 is the CIE pointer)

## Changes

- **`DebugFrameInputSection`** — parses input `.debug_frame` sections into CIE/FDE pieces
- **`DebugFrameSection`** — synthesizes the output section, keeping only live FDEs and their referenced CIEs
- **`combineDebugFrameSections`** — merges per-file input sections into the synthetic output
- Section is created lazily when any input has `.debug_frame` content

## Test plan

- [ ] Existing LLD test suite passes (`ninja check-lld`)
- [ ] Manual: link an MOS program with `-g --gc-sections` and confirm no relocation errors
- [ ] Manual: verify surviving `.debug_frame` records reference only live functions

## Motivation

Needed to close the debug-support gap on MOS. Without this, `-g` builds of
any non-trivial MOS program fail at link.

### DIFF
--- a/lld/ELF/InputSection.cpp
+++ b/lld/ELF/InputSection.cpp
@@ -18,12 +18,15 @@
 #include "lld/Common/DWARF.h"
 #include "llvm/Support/Compiler.h"
 #include "llvm/Support/Compression.h"
+#include "llvm/Support/Debug.h"
 #include "llvm/Support/Endian.h"
 #include "llvm/Support/LEB128.h"
 #include "llvm/Support/xxhash.h"
 #include <algorithm>
 #include <optional>
 #include <vector>
+
+#define DEBUG_TYPE "lld"
 
 using namespace llvm;
 using namespace llvm::ELF;
@@ -217,6 +220,21 @@ uint64_t SectionBase::getOffset(uint64_t offset) const {
     if (!es->content().empty())
       if (InputSection *isec = es->getParent())
         return isec->outSecOff + es->getParentOffset(offset);
+    return offset;
+  }
+  case DebugFrame: {
+    // Analogous to EHFrame's second path only. .debug_frame pieces (CIEs and
+    // FDEs) may be dropped when their referenced functions are GC'd, so an
+    // input offset does not equal the output offset -- use getParentOffset
+    // to walk the surviving pieces and produce the compacted output offset.
+    //
+    // The EHFrame case has a first path for crtbegin/crtbeginT boundary
+    // symbols; that does not apply here because .debug_frame is not
+    // SHF_ALLOC and no CRT code references its start.
+    const DebugFrameInputSection *dfs = cast<DebugFrameInputSection>(this);
+    if (!dfs->content().empty())
+      if (SyntheticSection *ss = dfs->getParent())
+        return ss->outSecOff + dfs->getParentOffset(offset);
     return offset;
   }
   case Merge:
@@ -1617,3 +1635,133 @@ template void EhInputSection::split<ELF32LE>();
 template void EhInputSection::split<ELF32BE>();
 template void EhInputSection::split<ELF64LE>();
 template void EhInputSection::split<ELF64BE>();
+
+template <class ELFT>
+DebugFrameInputSection::DebugFrameInputSection(ObjFile<ELFT> &f,
+                                               const typename ELFT::Shdr &header,
+                                               StringRef name)
+    : InputSectionBase(f, header, name, InputSectionBase::DebugFrame) {}
+
+SyntheticSection *DebugFrameInputSection::getParent() const {
+  return cast_or_null<SyntheticSection>(parent);
+}
+
+// Return the offset within the parent synthetic section for a given input
+// offset. Mirrors EhInputSection::getParentOffset: pieces (CIEs, FDEs) may
+// be dropped when the FDE's target function is GC'd, and the surviving
+// pieces are re-packed in DebugFrameSection::finalizeContents. A dropped
+// piece keeps its default outputOff of -1; report a piece-relative offset
+// in that case so that callers do not produce addresses that overlap
+// unrelated pieces in the compacted output.
+uint64_t DebugFrameInputSection::getParentOffset(uint64_t offset) const {
+  auto it = partition_point(
+      fdes, [=](EhSectionPiece p) { return p.inputOff <= offset; });
+  if (it == fdes.begin() || it[-1].inputOff + it[-1].size <= offset) {
+    it = partition_point(
+        cies, [=](EhSectionPiece p) { return p.inputOff <= offset; });
+    if (it == cies.begin()) // invalid piece
+      return offset;
+  }
+  if (it[-1].outputOff == -1) // dropped piece
+    return offset - it[-1].inputOff;
+  return it[-1].outputOff + (offset - it[-1].inputOff);
+}
+
+// .debug_frame is a sequence of CIE or FDE records, similar to .eh_frame
+// but with different CIE identification (0xFFFFFFFF instead of 0).
+template <class ELFT> void DebugFrameInputSection::split() {
+  LLVM_DEBUG(llvm::dbgs()
+             << "DebugFrameInputSection::split(): file=" << file->getName()
+             << " contentSize=" << content().size() << "\n");
+  const RelsOrRelas<ELFT> elfRels = relsOrRelas<ELFT>();
+  if (elfRels.areRelocsCrel())
+    preprocessRelocs<ELFT>(elfRels.crels);
+  else if (elfRels.areRelocsRel())
+    preprocessRelocs<ELFT>(elfRels.rels);
+  else
+    preprocessRelocs<ELFT>(elfRels.relas);
+  LLVM_DEBUG(llvm::dbgs() << "DebugFrameInputSection::split(): preprocessed "
+                          << rels.size() << " relocs\n");
+
+  // The loop below expects the relocations to be sorted by offset.
+  auto cmp = [](const Relocation &a, const Relocation &b) {
+    return a.offset < b.offset;
+  };
+  if (!llvm::is_sorted(rels, cmp))
+    llvm::stable_sort(rels, cmp);
+
+  ArrayRef<uint8_t> d = content();
+  const char *msg = nullptr;
+  unsigned relI = 0;
+  while (!d.empty()) {
+    if (d.size() < 4) {
+      msg = "CIE/FDE too small";
+      break;
+    }
+    uint64_t size = endian::read32<ELFT::Endianness>(d.data());
+    if (size == 0) // ZERO terminator
+      break;
+    uint32_t id = endian::read32<ELFT::Endianness>(d.data() + 4);
+    size += 4;
+    if (LLVM_UNLIKELY(size > d.size())) {
+      msg = size == UINT32_MAX + uint64_t(4)
+                ? "CIE/FDE too large"
+                : "CIE/FDE ends past the end of the section";
+      break;
+    }
+
+    // Find the first relocation that points to [off,off+size). Relocations
+    // have been sorted by r_offset.
+    const uint64_t off = d.data() - content().data();
+    while (relI != rels.size() && rels[relI].offset < off)
+      ++relI;
+    unsigned firstRel = -1;
+    if (relI != rels.size() && rels[relI].offset < off + size)
+      firstRel = relI;
+    // In .debug_frame, CIE is identified by id == 0xFFFFFFFF (not 0 like .eh_frame)
+    LLVM_DEBUG(llvm::dbgs()
+               << "DebugFrameInputSection::split(): record at " << off
+               << " size=" << size << " id=" << id << " is "
+               << (id == 0xFFFFFFFF ? "CIE" : "FDE")
+               << " firstRel=" << (int)firstRel << "\n");
+    (id == 0xFFFFFFFF ? cies : fdes).emplace_back(off, this, size, firstRel);
+    d = d.slice(size);
+  }
+  LLVM_DEBUG(llvm::dbgs() << "DebugFrameInputSection::split(): found "
+                          << cies.size() << " CIEs and " << fdes.size()
+                          << " FDEs\n");
+  if (msg)
+    Err(file->ctx) << "corrupted .debug_frame: " << msg << "\n>>> defined in "
+                   << getObjMsg(d.data() - content().data());
+}
+
+template <class ELFT, class RelTy>
+void DebugFrameInputSection::preprocessRelocs(Relocs<RelTy> elfRels) {
+  Ctx &ctx = file->ctx;
+  rels.reserve(elfRels.size());
+  for (auto rel : elfRels) {
+    uint64_t offset = rel.r_offset;
+    Symbol &sym = file->getSymbol(rel.getSymbol(ctx.arg.isMips64EL));
+    RelType type = rel.getType(ctx.arg.isMips64EL);
+    RelExpr expr = ctx.target->getRelExpr(type, sym, content().data() + offset);
+    int64_t addend =
+        RelTy::HasAddend
+            ? getAddend<ELFT>(rel)
+            : ctx.target->getImplicitAddend(content().data() + offset, type);
+    rels.push_back({expr, type, offset, addend, &sym});
+  }
+}
+
+template DebugFrameInputSection::DebugFrameInputSection(ObjFile<ELF32LE> &,
+                                                        const ELF32LE::Shdr &, StringRef);
+template DebugFrameInputSection::DebugFrameInputSection(ObjFile<ELF32BE> &,
+                                                        const ELF32BE::Shdr &, StringRef);
+template DebugFrameInputSection::DebugFrameInputSection(ObjFile<ELF64LE> &,
+                                                        const ELF64LE::Shdr &, StringRef);
+template DebugFrameInputSection::DebugFrameInputSection(ObjFile<ELF64BE> &,
+                                                        const ELF64BE::Shdr &, StringRef);
+
+template void DebugFrameInputSection::split<ELF32LE>();
+template void DebugFrameInputSection::split<ELF32BE>();
+template void DebugFrameInputSection::split<ELF64LE>();
+template void DebugFrameInputSection::split<ELF64BE>();
--- a/lld/ELF/Config.h
+++ b/lld/ELF/Config.h
@@ -46,6 +46,7 @@ class ELFFileBase;
 class SharedFile;
 class InputSectionBase;
 class EhInputSection;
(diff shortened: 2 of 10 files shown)


mysterymath:
This is kind of a surprising change, and my suspicion would be that we're holding something wrong. Why is this issue selective to MOS and not the other targets?

johnwbyrd:
Possibly because other targets assume that nothing can ever be positioned at the zero location in memory.  The null pointer is a valid memory location in MOS, and it should not necessarily be garbage collected.

johnwbyrd:
In other words, MOS uses --gc-sections by default, and CFI + GC creates dangling relocations when functions get collected.

johnwbyrd:
Digging into this further -- the failure mode isn't a link error, it's silent DWARF corruption. Constructed a minimal repro:

```
// dead_functions.c
int unused_a(int x) { return x * 2; }
int unused_b(int x) { return x + 3; }
int unused_c(int x, int y) { return x + y; }
int unused_d(void) { return 42; }
int used(int x) { return x + 1; }
int main(void) { return used(0); }
```

Compile with mos-clang -g -c -fno-lto → object has 6 .text.* sections (-ffunction-sections default) and 6 FDEs in .debug_frame, one per function.

Link with --gc-sections (MOS default). 4 unused functions get collected. Output .debug_frame:

```
FDE  pc=00000000...00000050
FDE  pc=00000000...0000004e
FDE  pc=00000000...00000077
FDE  pc=00000000...00000005
FDE  pc=00000000...00000054
FDE  pc=00000000...00000030
```

All 6 FDEs claim to describe code starting at 0x00000000. The relocations into GC'd sections got zeroed. Two functions actually survive; four of those FDEs are lies. No link error. No warning. A debugger reading this sees multiple overlapping unwind rules at address 0 -- behavior undefined.

Why MOS specifically hits this:

clang/lib/Driver/ToolChains/MOSToolchain.cpp:75 unconditionally passes --gc-sections to the linker (except for -r).
MOSMCAsmInfo sets ExceptionsType = ExceptionHandling::DwarfCFI (added by the DWARF numbering + CFI commit in this stack), so clang emits .debug_frame FDEs under -g.
Base MCAsmInfo.cpp defaults ExceptionsType to ExceptionHandling::None.

Why other targets don't complain, per target:

AVR: leaves ExceptionsType = None (base default). No .debug_frame FDEs are emitted, so no dangling relocations to corrupt.

MSP430: has the same combo — ExceptionsType = DwarfCFI (MSP430MCAsmInfo.cpp) plus --gc-sections default (MSP430.cpp:279). Would produce the same silently-corrupt .debug_frame under -g --gc-sections. No matching issue in llvm-project's tracker — probably nobody debugs MSP430 with -g --gc-sections at enough scale to notice.

RISC-V embedded, ARM Cortex-M: same class of target, same latent risk if .debug_frame is emitted with GC.
Upstream lld's .debug_frame handling today: grep -c debug_frame lld/ELF/SyntheticSections.cpp = 0. .eh_frame has 33 references worth of FDE-liveness plumbing. .debug_frame gets nothing — it's passed through as-is, and dead FDEs come along for the ride with their relocations zeroed.

This PR extends the existing .eh_frame FDE-liveness handling to .debug_frame. Same shape, same intent: drop FDEs whose functions were garbage-collected, so the output only describes code that's actually present. MSP430 (and any future DwarfCFI + --gc-sections target) benefits automatically.

mysterymath:
> Digging into this further -- the failure mode isn't a link error, it's silent DWARF corruption. Constructed a minimal repro:
> 
> ```
> // dead_functions.c
> int unused_a(int x) { return x * 2; }
> int unused_b(int x) { return x + 3; }
> int unused_c(int x, int y) { return x + y; }
> int unused_d(void) { return 42; }
> int used(int x) { return x + 1; }
> int main(void) { return used(0); }
> ```
> 
> Compile with mos-clang -g -c -fno-lto → object has 6 .text.* sections (-ffunction-sections default) and 6 FDEs in .debug_frame, one per function.
> 
> Link with --gc-sections (MOS default). 4 unused functions get collected. Output .debug_frame:
> 
> ```
> FDE  pc=00000000...00000050
> FDE  pc=00000000...0000004e
> FDE  pc=00000000...00000077
> FDE  pc=00000000...00000005
> FDE  pc=00000000...00000054
> FDE  pc=00000000...00000030
> ```
> 
> All 6 FDEs claim to describe code starting at 0x00000000. The relocations into GC'd sections got zeroed. Two functions actually survive; four of those FDEs are lies. No link error. No warning. A debugger reading this sees multiple overlapping unwind rules at address 0 -- behavior undefined.
> 
> Why MOS specifically hits this:
> 
> clang/lib/Driver/ToolChains/MOSToolchain.cpp:75 unconditionally passes --gc-sections to the linker (except for -r). MOSMCAsmInfo sets ExceptionsType = ExceptionHandling::DwarfCFI (added by the DWARF numbering + CFI commit in this stack), so clang emits .debug_frame FDEs under -g. Base MCAsmInfo.cpp defaults ExceptionsType to ExceptionHandling::None.
> 
> Why other targets don't complain, per target:
> 
> AVR: leaves ExceptionsType = None (base default). No .debug_frame FDEs are emitted, so no dangling relocations to corrupt.
> 
> MSP430: has the same combo — ExceptionsType = DwarfCFI (MSP430MCAsmInfo.cpp) plus --gc-sections default (MSP430.cpp:279). Would produce the same silently-corrupt .debug_frame under -g --gc-sections. No matching issue in llvm-project's tracker — probably nobody debugs MSP430 with -g --gc-sections at enough scale to notice.
> 
> RISC-V embedded, ARM Cortex-M: same class of target, same latent risk if .debug_frame is emitted with GC. Upstream lld's .debug_frame handling today: grep -c debug_frame lld/ELF/SyntheticSections.cpp = 0. .eh_frame has 33 references worth of FDE-liveness plumbing. .debug_frame gets nothing — it's passed through as-is, and dead FDEs come along for the ride with their relocations zeroed.
> 
> This PR extends the existing .eh_frame FDE-liveness handling to .debug_frame. Same shape, same intent: drop FDEs whose functions were garbage-collected, so the output only describes code that's actually present. MSP430 (and any future DwarfCFI + --gc-sections target) benefits automatically.

Ah, okay. I've actually had conversations with Roland about this;  this is supposed to produce DWARF tombstones. References to garbage collected functions in DWARF are to be rewritten to a sentinel value (in this case, 0) to indicate to the debugger that the referenced code didn't make it into the binary. The linker does this explicitly; see here: https://github.com/llvm-mos/llvm-mos/blob/9142aebae87d0bf6fc7c55b05a415f2c188b19f3/lld/ELF/InputSection.cpp#L1055

We may need to tell the debugger about this somehow, if it's causing problems with LLDB. But as far as I can tell tombstoning is already the blessed way of handling linker GC + DWARF. I don't think we should invent a new one unless we can't find a way to get the old one working.

EDIT: I had accidentally been looking at the wasm copy of this logic; rewrote the comment to the ELF one.

johnwbyrd:
Thanks for pointing to the tombstone mechanism -- I dug into it and want to share what I found, because I think there's a gap here worth walking through.

You're right that lld does implement tombstoning: `InputSection.cpp:1064-1156` writes 0 (or -1 or 1 depending on section) into relocations whose target symbols were discarded by GC. That mechanism works cleanly for sections where the DWARF spec codified the convention and consumers implement matching skip logic:

- `.debug_line` -- LLVM reader at `DWARFDebugLine.cpp:859-970` calls `dwarf::computeTombstoneAddress()` and skips tombstoned rows.
- `.debug_ranges` / `.debug_rnglists` -- same helper, at `DWARFDebugRangeList.cpp:23,93` and `DWARFDebugRnglists.cpp:102,205`.
- `.debug_addr` (via `DW_AT_low_pc`) -- same helper at `DWARFDie.cpp:212,423`.

For those sections, tombstoning is the blessed way because the round-trip closes: producer writes the sentinel, consumer skips it.

For `.debug_frame`, the round-trip doesn't close. Consumer side:

- `llvm/lib/DebugInfo/DWARF/DWARFDebugFrame.cpp` -- `grep -c tombstone` = 0. `computeTombstoneAddress` is never called. `InitialLocation` is read at line 379 and passed straight into the FDE constructor.
- `lldb/source/Symbol/DWARFCallFrameInfo.cpp` -- same story. Lines 614-622 read `range_base` and construct `AddressRange` unconditionally.

Spec side: DWARF 5 section 7.29 codifies the base-address-specifier / tombstone convention for the ranges/addresses family. Section 6.4.1 (Call Frame Information) does not extend it to CFI. There are 33 `DW_CFA_*` opcodes in `BinaryFormat/Dwarf.def`; none carry tombstone or discarded semantics. Roland's convention is real and reasonable, but it hasn't landed in a codified form that consumers follow.

I re-ran the repro just now with both linkers -- upstream/main at `9142aebae87d` vs current llvm-mos/llvm-mos main (which already carries this patch):

```
                  upstream lld           llvm-mos lld (with this PR)
Link result:      exit 0, no warning     exit 0, no warning
FDEs in output:   6                      2
FDEs at pc=0x0:   4                      0
```

Upstream output:
```
00000028  FDE  pc=00000000...00000050   -- unused_a tombstoned
0000005c  FDE  pc=00000000...0000004e   -- unused_b tombstoned
00000090  FDE  pc=00000000...00000077   -- unused_c tombstoned
000000c4  FDE  pc=00000000...00000005   -- unused_d tombstoned
000000d4  FDE  pc=000100b4...00010108   -- used (real)
00000108  FDE  pc=00010108...00010138   -- main (real)
```

With the patch: only `used` and `main`, both at their real addresses.

A debugger reading the upstream output sees five FDEs claiming code at address 0 (four tombstones plus the CIE marker convention). None of them are flagged as tombstones because no consumer implements the recognition. On MOS specifically this matters an extra amount because address 0 is valid code space, so we can't even fall back to "the debugger will notice 0 is obviously bogus" -- a legitimate FDE at 0 is possible in principle.

Two paths to close the gap:

1. Add tombstone recognition to `DWARFDebugFrame.cpp` and `DWARFCallFrameInfo.cpp` upstream, formalize the convention for CFI in the DWARF spec, and wait for it to propagate through GDB, LLDB, and every other consumer.
2. Drop tombstoned FDEs at link time, mirroring the existing `.eh_frame` pattern, so no consumer has to change.

This PR is option 2, which I chose because it's the same shape as the `.eh_frame` handling already in this file, doesn't require a spec change, and works with every consumer today. MSP430 and any future DwarfCFI-plus-`--gc-sections` target inherits the fix automatically.

Would you be willing to take the patch?

mysterymath:
Thanks for your analysis; I've followed up on the details on my end, and it looks like this is an existing gap upstream, and a PR very much like this was proposed as a solution. I've started reviewing the contents in detail; I'll get back to you once I'm further underway.

lld/ELF/InputSection.cpp
```diff
@@ -219,6 +220,14 @@ uint64_t SectionBase::getOffset(uint64_t offset) const {
         return isec->outSecOff + es->getParentOffset(offset);
     return offset;
   }
+  case DebugFrame: {
+    // Similar to EHFrame but for .debug_frame sections.
```
mysterymath: From the above it appears that there is a very restrictive set of scenarios where `getOffset` might be called on `EhFrame`. It also appears that the semantics are tailored for those uses. This seems like it should be explicitly mentioned; it prima facie can't be the same set of cases, because EHFrame is an alloc section in the actual binary, and debug_frame is not. Accordingly, we should make sure that this code is actually needed, and if so, make sure the semantics are still appropriately tailored. I'm not *necessarily* saying that anything is *wrong* with this code, just that there's enough different we should be explicit about the extents of the analogy to EHFrame.

mysterymath (commented):
This change pretty substantially alters the `.debug_frame` handling under `--gc-sections`. This means that it needs tests: both for the garbage collection behavior in question, and at least one smoke test to verify that its behavior is logically equivalent to the prior when GC does not actually occur.

lld/ELF/InputSection.cpp
```diff
@@ -1617,3 +1626,112 @@ template void EhInputSection::split<ELF32LE>();
...
+                                               StringRef name)
+    : InputSectionBase(f, header, name, InputSectionBase::DebugFrame) {}
+
+SyntheticSection *DebugFrameInputSection::getParent() const {
+  return cast_or_null<SyntheticSection>(parent);
+}
+
+// .debug_frame is a sequence of CIE or FDE records, similar to .eh_frame
+// but with different CIE identification (0xFFFFFFFF instead of 0).
+template <class ELFT> void DebugFrameInputSection::split() {
```
mysterymath, line 1672: Without having looked *too* much into the details, it seems very likely that, by virtue of the analogies used in describing this patch, that the behavior of the linker WRT `.eh_frame` and `.debug_frame` are, well, very analogous. Could we have a single abstraction that handles both, with specific carve-outs for the differences between them?

lld/ELF/InputSection.cpp
johnwbyrd: You're right that the two cases in the EHFrame path aren't both applicable. The crtbegin/crtbeginT boundary-symbol path doesn't translate here -- `.debug_frame` is not `SHF_ALLOC` and no CRT code references its start. The `copyRelocations`/`--emit-relocs` path does apply, and the original code was wrong for it: it used the raw `offset` even though pieces are dropped and re-packed in `finalizeContents`, so an input offset into a dropped or shifted piece would yield a wrong or overlapping output offset.

Amended: added `DebugFrameInputSection::getParentOffset(uint64_t)` mirroring `EhInputSection::getParentOffset` (walks `fdes` then `cies`, maps `inputOff -> outputOff`, returns a piece-relative offset for dropped pieces so callers don't index into unrelated live pieces), and changed this case to `ss->outSecOff + dfs->getParentOffset(offset)`. Expanded the case comment to state which EHFrame paths translate and which don't.

lld/ELF/InputSection.cpp
johnwbyrd: The analogy is real, and I considered it. The carve-outs are meaningful: CIE marker constant (`0` vs `0xFFFFFFFF`), CIE-pointer semantics (relative in FDE header vs absolute), FDE function-address offset (0 vs 8), section flags (`SHF_ALLOC` vs not), partition membership (`.eh_frame` has partitions, `.debug_frame` doesn't), cross-file CIE dedup (`.eh_frame` merges by content, `.debug_frame` here does not), personality/LSDA reference handling (only `.eh_frame`), and write-time fixups (CIE-pointer patch for `.debug_frame`, PLT edge cases for `.eh_frame`). Wrapping those in a Policy<> template or a virtual-fn base without regressing the `.eh_frame` hot path would be a non-trivial refactor.

Given the risk profile -- `.eh_frame` is exercised by every C++ program's exception path -- I'd prefer to land this as its own extension of the pattern and do the unification as a follow-up that can be reviewed on its own merits. Structurally the two paths stay close, so a later merge has less rebase burden. If you'd rather I attempt the unification here, say so and I'll prototype it; I just want to flag that touching `EhFrameSection` is the more delicate surgery of the two.

johnwbyrd:
Added `lld/test/ELF/gc-sections-debug-frame.s` and `lld/test/ELF/gc-sections-debug-frame-multifile.s`, mirroring `gc-sections-eh.s`:

- **Single-file GC**: `dead_fn` collected (readobj), one live FDE with a non-zero PC (dwarfdump), no stale FDE at address 0.
- **Single-file no-GC**: both symbols and both FDEs present -- smoke test for pass-through equivalence.
- **Multi-file GC**: two objects, each with a live + dead function. Asserts each surviving FDE references its own file's CIE via FileCheck capture-and-reuse (`[[CIE_A]]` / `[[CIE_B]]`).

While writing the multi-file test I also fixed two pre-existing correctness bugs in this PR that the test exposed:

- `cieInputToOutput` in `writeTo` was keyed by input-offset alone. Every `.o` typically emits its CIE at input-offset 0, so with multiple input files the map key collided and every FDE ended up referencing whichever file's CIE was inserted last. Fixed by keying on `(InputSectionBase*, uint64_t)`.
- `finalizeContents` and `writeTo` used hardcoded `endian::read32le` / `write32le` for id and CIE-pointer bytes. On BE targets the read/write would byte-swap the value, corrupting the output. Fixed by splitting `livePieces` into `liveCies`/`liveFdes` (classification is already known at `addRecords<E>` time) so downstream code never re-reads id bytes, and by templating `writeTo` on endianness via a `writeToImpl<E>` mirroring the `addRecords<E>` dispatch pattern.

Also converted the exploratory `DEBUG_WITH_TYPE("debug-frame", ...)` logs (28 sites) to `LLVM_DEBUG(llvm::dbgs() << ...)` to match `lld/ELF/Writer.cpp`'s local style, added `#define DEBUG_TYPE "lld"` in the three files that were missing it, and sorted the new `Debug.h` includes into alphabetical position.

Full `check-lld-elf` passes (2036/2036, no regressions).

mysterymath:
I ended up talking about this with some knowledgeable folks; I think the reason this is breaking for us specifically is that LLDB  checks if anything is mapped into memory for 0, and if so, ceases to treat is as a tombstone. It doesn't have this behavior for the dedicated tombstone value of -1, and there's a LLD flag to make it emit that value. (I think also for `.debug_frame`? Haven't looked into it in detail.)

That being said, I'd like to move forward with this patch, since it seems to be something that folks would actually like upstream, since it's more elegant. @johnwbyrd, would you be open to opening this up as an upstream PR once we get it into a good shape here? It would be something that would help with embedded debugging generally, since ARM apparently does map things to zero too.

mysterymath:
I'd also be willing to drive that process if you'd prefer, with attribution to your work.

mysterymath (approved):
Alright, there's a lot I didn't verify, but I don't think exhaustive correctness WRT all the stuff LLD supports is necessary for llvm-mos yet. Accordingly, this LGTM for our purposes. Thanks!

PR #568, opened 2026-07-06
feat(MC): add first-class CFI expression instruction support

johnwbyrd:
## Summary

Add `OpDefCfaExpression`, `OpExpression`, and `OpValExpression` variants to
`MCCFIInstruction` for targets that require DWARF expressions in CFI
directives.

## Background

DWARF CFI supports three expression-based opcodes:

- `DW_CFA_def_cfa_expression` — define CFA via a DWARF expression
- `DW_CFA_expression` — express a register's location as an expression
- `DW_CFA_val_expression` — express a register's value as an expression

LLVM's `MCCFIInstruction` had no first-class representation for these,
forcing targets to emit raw `.cfi_escape` byte sequences. Targets with
dual-stack or otherwise non-constant call frames need real expressions to
describe the return address, and `.cfi_escape` loses type information and
cannot be validated.

## Changes

- **`MCDwarf`** — three new `OpKind` values and constructor helpers
- **`AsmPrinterDwarf`** — emit the new instructions via the assembler streamer
- **`CFIInstrInserter`** — treat expressions as opaque for the RA-analysis passes it runs
- **`MIRParser`/`MIParser`** — parse and print the three new CFI forms in MIR
- **`DWARFCFIChecker`** — emit a diagnostic when expression forms appear in verified CFI (checker doesn't interpret expressions)

## Test plan

- [ ] Existing MC/MIR tests pass (`ninja check-llvm-unit check-llvm-mc`)
- [ ] MIR round-trip test for each new opcode

## Motivation

Prerequisite for CFI on architectures whose call frame cannot be described
with the constant-offset opcodes. MOS uses this for its dual-stack layout,
but the mechanism is generic — expression forms are already part of the
DWARF spec and any target with non-standard frames can use them.


mysterymath:
Similarly to the previous commit, this adds a lot of mechanism; the stated purposes it to support MOS's dual-stack layout. What about that is overly-cumbersome or impossible to represent using the constructs that already exist in the AsmPrinter/MC layer? Are we certain that none of the other targets deal with similar constraints? If they do, what do they do here?

johnwbyrd:
I assume that LLVM hasn't had to emit code that deals with multiple stacks in this way before.  lldb fortunately parses the CFI expressions correctly, so no changes were needed there.  We require expression instruction support in order to create expressions that deal with 1) multiple stacks as well as 2) multiple debuggers that see the S register in multiple ways.  Upstream should be happy with this particular code, as it's missing and will be useful to future platforms that need it.

johnwbyrd:
Helpful for intuition: https://www.llvm-mos.org/wiki/DWARF_implementation_guide

johnwbyrd:
On upstream, MCCFIInstruction has no first-class expression forms. Every target that needs one (X86 in X86FrameLowering.cpp × 4 sites, RISC-V in RISCVFrameLowering.cpp × 2 sites, AArch64 in AArch64InstrInfo.cpp × 2 sites — 8 hand-rolled call sites) falls back to MCCFIInstruction::createEscape and manually builds the DWARF opcode + ULEB128 length + payload themselves.  Rather than adding a fourth copy of this skeleton, we let the primitive be first-class, which lets the structures be tracked smartly through IR.

mysterymath:
> On upstream, MCCFIInstruction has no first-class expression forms. Every target that needs one (X86 in X86FrameLowering.cpp × 4 sites, RISC-V in RISCVFrameLowering.cpp × 2 sites, AArch64 in AArch64InstrInfo.cpp × 2 sites — 8 hand-rolled call sites) falls back to MCCFIInstruction::createEscape and manually builds the DWARF opcode + ULEB128 length + payload themselves. Rather than adding a fourth copy of this skeleton, we let the primitive be first-class, which lets the structures be tracked smartly through IR.

I had Claude look into this, and it really seems like the RISC-V model is the benchmark to follow. A helper to produce these expressions should be pretty straightforward and tiny compared to the mechanism introduced here:

```
#include "llvm/BinaryFormat/Dwarf.h"
#include "llvm/Support/LEB128.h"

// DW_CFA_def_cfa_expression <len> <expr bytes>
static MCCFIInstruction createDefCFAExpression(ArrayRef<uint8_t> Expr,
                                               const Twine &Comment) {
  SmallString<64> CFI;
  raw_svector_ostream OS(CFI);
  OS << uint8_t(dwarf::DW_CFA_def_cfa_expression);
  encodeULEB128(Expr.size(), OS);
  OS << StringRef(reinterpret_cast<const char *>(Expr.data()), Expr.size());
  return MCCFIInstruction::createEscape(nullptr, OS.str(), SMLoc(),
                                        Comment.str());
}

// DW_CFA_expression / DW_CFA_val_expression <reg> <len> <expr bytes>
static MCCFIInstruction createRegExpression(uint8_t CFAOpcode,
                                            unsigned DwarfReg,
                                            ArrayRef<uint8_t> Expr,
                                            const Twine &Comment) {
  SmallString<64> CFI;
  raw_svector_ostream OS(CFI);
  OS << CFAOpcode;
  encodeULEB128(DwarfReg, OS);
  encodeULEB128(Expr.size(), OS);
  OS << StringRef(reinterpret_cast<const char *>(Expr.data()), Expr.size());
  return MCCFIInstruction::createEscape(nullptr, OS.str(), SMLoc(),
                                        Comment.str());
}
```

If we want to introduce general mechanism for this kind of thing, we can, but it's the kind of thing that should loop the existing customers in, and that would tend to mean making this an upstream PR that brings everyone else along. That would allow enough eyes on the problem to ensure a good design; otherwise, it seems better to do a llvm-mos specific usage of existing substrates. We do make exceptions for cases where there's simply *no good way* to use the existing stuff to achieve good 6502 outcomes, but that doesn't seem to be the case here.\

EDIT: And helpers/approaches like this are small enough to be added inline with the code that uses them.

johnwbyrd:
Superseded by #600.

After discussion, adding first-class `OpExpression` / `OpValExpression` / `OpDefCfaExpression` variants to upstream `MCCFIInstruction` grows cross-target LLVM mechanism without upstream buy-in (@mysterymath's objection). #600 reaches the same outcome via MOS-only `mos::create{Expression,ValExpression,DefCfaExpression}` helpers in `MOSCFIExprHelpers.h`, which wrap `MCCFIInstruction::createEscape` with first-class-shaped signatures. If upstream ever grows real constructors, migration is a body swap on three helper functions with no call-site churn.

Closing in favor of #600.

PR #569, opened 2026-07-06
feat(AsmPrinter): allow targets to defer handler initialization

johnwbyrd:
## Summary

Add `shouldCallBeginModule()` virtual and `callBeginModule()` method to
`AsmPrinter`, letting targets defer debug/EH handler initialization until
after late passes that mutate the global list.

## Background

`DwarfDebug::beginModule` iterates `Module::globals()` and materializes
DIE entries with relocations. If a target subsequently runs a pass that
deletes globals — MOS runs `MOSInternalize` which drops unreferenced
globals — the relocations in `.debug_addr` become dangling and the linker
emits errors against removed symbols.

## Changes

- **`AsmPrinter::shouldCallBeginModule()`** — new virtual, default `true` preserves current behavior
- **`AsmPrinter::callBeginModule()`** — public entry point targets can call after their late passes
- `AsmPrinter::doInitialization` skips the eager call when `shouldCallBeginModule()` returns `false`

## Test plan

- [ ] All existing AsmPrinter tests pass unchanged (default path)
- [ ] MOS target uses this to gate `beginModule` until after `MOSInternalize`

## Motivation

Prerequisite for reliable `-g` builds on MOS, where `MOSInternalize` removes
globals after `beginModule` would otherwise have committed relocations
against them. The hook is target-generic — the default
(`shouldCallBeginModule() = true`) leaves every existing target unchanged.

### DIFF
--- a/llvm/include/llvm/CodeGen/AsmPrinter.h
+++ b/llvm/include/llvm/CodeGen/AsmPrinter.h
@@ -318,6 +318,29 @@ public:
     return nullptr;
   }
 
+  /// Returns true if beginModule should be called during doInitialization.
+  ///
+  /// DwarfDebug::beginModule iterates over all GlobalVariables in the
+  /// module and creates debug info entries (DIEs) for them. Some
+  /// targets (like MOS) run GlobalDCE after AsmPrinter::doInitialization,
+  /// which can delete GlobalVariables that debug info already references.
+  /// This causes "undefined temporary symbol" linker errors because the
+  /// .debug_addr section contains relocations to symbols that no longer exist.
+  ///
+  /// Targets that run GlobalDCE after codegen starts can override this to
+  /// return false and call beginModule later via callBeginModule(), after
+  /// GlobalDCE has finished. This ensures debug info only references globals
+  /// that survive dead code elimination.
+  virtual bool shouldCallBeginModule() const { return true; }
+
+  /// Manually trigger beginModule for all debug/EH handlers.
+  ///
+  /// This is called automatically during doInitialization if
+  /// shouldCallBeginModule() returns true. Targets that override
+  /// shouldCallBeginModule() to return false, must call this manually
+  /// after any passes that may delete GlobalVariables have completed.
+  void callBeginModule(Module *M);
+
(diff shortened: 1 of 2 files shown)


mysterymath:
*Sigh*, I was afraid of something like this happening. I've received warnings from upstream folks that it's *not* generally safe to modify the IR late in codegen like we do in `MOSInternalize`; this might be an instance of this. I'll want to take a detailed look at this; the right solution might be to move or modify `MOSInternalize` to be safer.

johnwbyrd:
That's fine -- just be aware that I consider this to be likely that this is gating source-level debugging.

mysterymath:
Ah, after I did a deep dive into the issue and prepared an alternative fix, the candidate fix actually... wasn't needed to make the test pass. It looks like this behavior was moved from `beginModule()` to `endModule()` in https://github.com/llvm/llvm-project/commit/63074da25d16376d4a5b27661d891623bf797fd3, back in May, so it should all happen after the final IR is available now. Are you sure this is still an issue?

johnwbyrd:
You're right — I walked through the current `DwarfDebug.cpp` and 63074da25d16 does move the `M->globals()` iteration out of `beginModule()` into `endModule()`. Since `endModule()` runs from `AsmPrinter::doFinalization()` (well after `MOSInternalize` has finished its `GlobalDCE`), the dangling `.debug_addr` relocations can't happen anymore. The `shouldCallBeginModule()` hook is solving a problem that no longer exists.

I verified this empirically on a fresh branch (`feature/debug/v38`) with both the AsmPrinter hook and the MOSAsmPrinter override removed. Full LLVM+clang+lld+lldb build is clean. Repro case (a static function whose only reference to a private const is `__func__`, guaranteed to be DCE'd because `main` doesn't call it, compiled `--target=mos -g -O2`):

- `llvm-dwarfdump --verify` — "No errors."
- `dead_helper` and `__func__` both gone from the symbol table (GlobalDCE did its job)
- `.debug_info` contains only `main`'s DIE — no stale entry for the DCE'd function
- `.rela.debug_addr` has a single entry pointing at `.text.main + 0` — nothing dangling

Closing this PR and dropping the corresponding override from `MOSAsmPrinter`. Thanks for the deep dive.

PR #570, opened 2026-07-06
feat(GlobalISel): implement debug info salvaging for G_MERGE_VALUES

johnwbyrd:
## Summary

Salvage debug info for `G_MERGE_VALUES` by rewriting affected `DBG_VALUE`s
to reference each source operand with a `DW_OP_LLVM_fragment` expression
covering the appropriate byte range. Call `salvageDebugInfo()` before
erasing instructions in `eraseInstrs()`.

## Background

When `G_MERGE_VALUES` is eliminated (dead-code, folded, or merged into a
larger instruction), any `DBG_VALUE` referencing its result becomes invalid
and the debugger loses visibility of the merged variable.

On targets like MOS, **every** type larger than 8 bits is split via
`G_MERGE_VALUES`/`G_UNMERGE_VALUES`. Without salvaging, virtually no
`i16`/`i32`/`i64` variable is inspectable at all optimization levels above
`-O0`.

## Changes

- **`CodeGenCommonISel::salvageDebugInfoForGMerge`** — new helper that constructs the fragment DIExpressions
- **`Utils::eraseInstrs`** — invoke the salvager before removing each instruction

## Test plan

- [ ] Existing GlobalISel debug tests pass (`ninja check-llvm`)
- [ ] `x86-calllowering-dbg-trunc.ll` — golden test now checks salvaged fragments (updated in this commit)
- [ ] Manual: MOS `-g -O2` build shows locations for split integer variables

## Motivation

Needed on MOS to make wide-integer variables inspectable above `-O0`, but
applies generically to any target that lowers wide types via
`G_MERGE_VALUES`/`G_UNMERGE_VALUES`.

### DIFF
--- a/llvm/lib/CodeGen/CodeGenCommonISel.cpp
+++ b/llvm/lib/CodeGen/CodeGenCommonISel.cpp
@@ -15,6 +15,8 @@
 #include "llvm/Analysis/BranchProbabilityInfo.h"
 #include "llvm/CodeGen/MachineBasicBlock.h"
 #include "llvm/CodeGen/MachineFunction.h"
+#include "llvm/CodeGen/MachineInstrBuilder.h"
+#include "llvm/CodeGen/MachineRegisterInfo.h"
 #include "llvm/CodeGen/TargetInstrInfo.h"
 #include "llvm/CodeGen/TargetOpcodes.h"
 #include "llvm/IR/DebugInfoMetadata.h"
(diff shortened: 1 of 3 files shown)


llvm/lib/CodeGen/CodeGenCommonISel.cpp
mysterymath, line 352: There appears to be a lot of commonality between the above broken-out function and this one. It seems desirable to have a single loop handling users with cases for various types of instruction; is that not feasible for some reason? If so, why?

It seems like it might be because the above produces multiple DBG_VALUE expressions instead of just one. Could we generalize the below to handle multiples instead of splitting out a copy of the whole routine just for G_MERGE_VALUES?

PR #571, opened 2026-07-06
feat(lldb): ABI extension points for custom register contexts, plus MOS ABI plugin

johnwbyrd:
## Summary

Add three extension points to `lldb::ABI` so targets with non-standard
register models can plug in their own register context and canonical
register info:

- `CreateRegisterContextForThread()` — provide a target-specific context
- `ProvidesRegisterInfoOverride()` — signal that the ABI supplies custom register handling
- `GetCanonicalRegisterInfo()` — post-process register definitions returned by a GDB stub

`ThreadGDBRemote` consults the ABI before creating the default
`GDBRemoteRegisterContext`. `ProcessGDBRemote` applies canonical register
info when the ABI provides it.

## Background

The GDB remote protocol reports only registers that the target's stub knows
about. On MOS, the imaginary-register file (`rc0`-`rc255`, `rs0`-`rs127`)
lives in zero-page memory and no stub reports these — but LLDB needs to
expose them to the user for debugging.

Rather than adding MOS-specific hooks throughout LLDB core, this PR adds
general extension points to the ABI class that any target can use.
Existing targets (default `ABI::CreateRegisterContextForThread()` returns
`nullptr`) get identical behavior.

## Changes

- **`lldb/include/lldb/Target/ABI.h`** — three new virtuals with default no-op implementations
- **`ThreadGDBRemote`** — check ABI before falling back to default context
- **`ProcessGDBRemote`** — merge canonical register info from ABI when available

## Test plan

- [ ] Existing LLDB test suite passes (`ninja check-lldb`)
- [ ] Follow-up MOS ABI plugin uses these hooks (separate PR)

## Motivation

The hook is target-generic — every existing ABI subclass gets the default
no-op implementations. Sets up the mechanism that a forthcoming MOS LLDB
ABI plugin will use to expose imaginary registers to the debugger.

### DIFF
--- a/lldb/include/lldb/Target/ABI.h
+++ b/lldb/include/lldb/Target/ABI.h
@@ -152,6 +152,21 @@ public:
 
   static lldb::ABISP FindPlugin(lldb::ProcessSP process_sp, const ArchSpec &arch);
 
+  virtual lldb::RegisterContextSP
+  CreateRegisterContextForThread(
+      lldb_private::Thread &thread,
+      uint32_t concrete_frame_idx) const {
+    // Default: return nullptr, meaning use the default RegisterContext
+    return nullptr;
+  }
+
+  // Returns true if the ABI is prepared to provide authoritative register info for override.
+  virtual bool ProvidesRegisterInfoOverride() const { return false; }
+
+  // If opted in, returns a fully-populated Register for the given name, or std::nullopt if not recognized.
+  virtual std::optional<lldb_private::DynamicRegisterInfo::Register>
+  GetCanonicalRegisterInfo(llvm::StringRef name) const { return std::nullopt; }
+
 protected:
   ABI(lldb::ProcessSP process_sp, std::unique_ptr<llvm::MCRegisterInfo> info_up)
       : m_process_wp(process_sp), m_mc_register_info_up(std::move(info_up)) {
--- a/lldb/include/lldb/Target/ABI.h
+++ b/lldb/include/lldb/Target/ABI.h
@@ -179,6 +179,22 @@ public:
     return std::nullopt;
   }
 
+  virtual lldb::RegisterContextSP
+  CreateRegisterContextForThread(
+      lldb_private::Thread &thread,
+      uint32_t concrete_frame_idx) const {
+    // Default: return nullptr, meaning use the default RegisterContext
+    return nullptr;
+  }
+
+  // Returns true if the ABI is prepared to provide authoritative register info for override.
+  virtual bool ProvidesRegisterInfoOverride() const { return false; }
+
+  // If opted in, returns a fully-populated Register for the given name, or std::nullopt if not recognized.
+  virtual std::optional<lldb_private::DynamicRegisterInfo::Register>
+  GetCanonicalRegisterInfo(llvm::StringRef name) const { return std::nullopt; }
+
+
 protected:
   ABI(lldb::ProcessSP process_sp, std::unique_ptr<llvm::MCRegisterInfo> info_up)
       : m_process_wp(process_sp), m_mc_register_info_up(std::move(info_up)) {
--- a/lldb/source/Plugins/ABI/MOS/ABISysV_mos.cpp
+++ b/lldb/source/Plugins/ABI/MOS/ABISysV_mos.cpp
@@ -1,0 +1,326 @@
+//===-- ABISysV_mos.cpp -----------------------------------------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "ABISysV_mos.h"
+#include "MOSImaginaryRegisters.h"
+#include "MOSRegisterContext.h"
+
+#include "lldb/Core/PluginManager.h"
+#include "lldb/Core/Value.h"
+#include "lldb/Symbol/UnwindPlan.h"
+#include "lldb/Target/Process.h"
+#include "lldb/Target/Thread.h"
+#include "lldb/Utility/LLDBLog.h"
+#include "lldb/Utility/Log.h"
+
+#include "Plugins/Process/gdb-remote/ThreadGDBRemote.h"
+
+#include "llvm/BinaryFormat/Dwarf.h"
+#include "llvm/TargetParser/Triple.h"
+
+using namespace lldb;
+using namespace lldb_private;
+
+LLDB_PLUGIN_DEFINE_ADV(ABISysV_mos, ArchitectureMOS)
+
+//------------------------------------------------------------------
+// Helper to look up DWARF register number by name via MCRegisterInfo.
+// This is DRY - the actual DWARF numbers come from MOSRegisterInfo.td
+// through the MCRegisterInfo generated at LLVM build time.
+//------------------------------------------------------------------
+static int GetDwarfRegNumByName(llvm::MCRegisterInfo *mc_info,
+                                llvm::StringRef name) {
+  if (!mc_info)
+    return -1;
+
+  // MCRegisterInfo stores names in uppercase
+  std::string mc_name = name.upper();
+
+  for (unsigned reg = 0; reg < mc_info->getNumRegs(); ++reg) {
+    if (mc_info->getName(reg) == mc_name) {
+      return mc_info->getDwarfRegNum(reg, /*isEH=*/false);
+    }
+  }
+  return -1;
+}
+
+//------------------------------------------------------------------
+// Static Functions
+//------------------------------------------------------------------
+
+void ABISysV_mos::Initialize() {
+  PluginManager::RegisterPlugin(GetPluginNameStatic(),
+                                "System V ABI for MOS targets", CreateInstance);
+}
+
+void ABISysV_mos::Terminate() {
+  MOSImaginaryRegisters::ClearCache();
+  PluginManager::UnregisterPlugin(CreateInstance);
+}
+
+ABISP ABISysV_mos::CreateInstance(lldb::ProcessSP process_sp,
+                                  const ArchSpec &arch) {
+  if (arch.GetTriple().getArch() == llvm::Triple::mos) {
+    return ABISP(
+        new ABISysV_mos(std::move(process_sp), MakeMCRegisterInfo(arch)));
+  }
+  return ABISP();
+}
+
+//------------------------------------------------------------------
+// MCBasedABI interface
+//------------------------------------------------------------------
+
+uint32_t ABISysV_mos::GetGenericNum(llvm::StringRef name) {
+  // Map register names to generic register numbers
+  if (name.equals_insensitive("pc"))
+    return LLDB_REGNUM_GENERIC_PC;
+  if (name.equals_insensitive("sp") || name.equals_insensitive("s"))
+    return LLDB_REGNUM_GENERIC_SP;
+  if (name.equals_insensitive("p"))
+    return LLDB_REGNUM_GENERIC_FLAGS;
+  // RS0 is the soft stack pointer (frame pointer)
+  if (name.equals_insensitive("rs0"))
+    return LLDB_REGNUM_GENERIC_FP;
+
+  return LLDB_INVALID_REGNUM;
+}
+
+void ABISysV_mos::AugmentRegisterInfo(
+    std::vector<DynamicRegisterInfo::Register> &regs) {
+  // First, call base class to handle standard DWARF/EH frame numbers
+  MCBasedABI::AugmentRegisterInfo(regs);
+
+  // Add imaginary registers if we have a main module
+  ModuleSP main_module = GetMainModule();
+  if (main_module && m_mc_register_info_up) {
+    MOSImaginaryRegisters &imag_regs = MOSImaginaryRegisters::GetOrCreate(
+        *main_module, *m_mc_register_info_up);
+    if (imag_regs.HasImaginaryRegisters()) {
+      imag_regs.AddToRegisterList(regs, *m_mc_register_info_up);
+
+      LLDB_LOG(GetLog(LLDBLog::Expressions),
+               "ABISysV_mos: Added {0} RC and {1} RS imaginary registers",
+               imag_regs.GetNumRCRegisters(), imag_regs.GetNumRSRegisters());
+    }
+  }
+}
+
+//------------------------------------------------------------------
+// ABI interface - trivial call support (not implemented for 6502)
+//------------------------------------------------------------------
+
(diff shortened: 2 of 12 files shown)


mysterymath:
I had requested smaller commits for the LLDB work, and I wanted to thank you for obliging. But, in this case, this commit is actually a bit *too* small to effectively review.

An API serves two parties: its implementation and its callers. The caller part is trivial here, so it's really the implementation and effects that matter. But, this change has no information about that, and I'd need to go digging into both in the external repo to review. Accordingly, it would be very preferable to bundle these interfaces with the changes that introduce the actual functional changes that require them; the extra 36 lines won't make it any more difficult, and actually, far to the contrary.

Leaving open, since you could repurpose this PR for this.

johnwbyrd:
Repurposed the PR as you suggested. Force-pushed the branch and folded in the caller as a second commit.

The PR now contains two commits:

1. **feat(lldb): add ABI extension points for custom register contexts** — the three virtual methods (`CreateRegisterContextForThread`, `ProvidesRegisterInfoOverride`, `GetCanonicalRegisterInfo`) plus the GDB-remote plumbing. Same as before.
2. **feat(lldb): add MOS ABI plugin for debugging support** — the first user of those extension points. `ABISysV_mos` + `MOSImaginaryRegisters` + `MOSRegisterContext`, entirely under `lldb/source/Plugins/ABI/MOS/`. ~1014 LoC. Was previously #602, now closed and folded here.

The two commits are kept separate rather than squashed so the API surface stays reviewable as a distinct hunk, but the caller is right next to it. Also updated the title to reflect combined scope.

Pairs with #601 (compiler-side DWARF/CFI) — together they're the "LLDB understands MOS" story, but the two are code-wise independent and can land in either order.

lldb/source/Plugins/Process/gdb-remote/ProcessGDBRemote.cpp
```diff
@@ -5613,11 +5613,17 @@ void ProcessGDBRemote::AddRemoteRegisters(
...
+  // ABI override for register info, if enabled
+  if (ABISP abi_sp = ABI::FindPlugin(shared_from_this(), arch_to_use)) {
+    if (abi_sp->ProvidesRegisterInfoOverride()) {
+      for (auto &reg : registers) {
+        auto canonical = abi_sp->GetCanonicalRegisterInfo(reg.name.GetStringRef());
+        if (canonical)
+          reg = *canonical;
+      }
+    }
     abi_sp->AugmentRegisterInfo(registers);
```
mysterymath: `AugmentRegisterInfo` already takes the vector of registers mutably, so it's the blessed way for a target to manipulate the list of registers. AArch64 and X86 already use this interface to add supplementary registers. Accordingly, it seems like these two additional ABI interfaces are redundant.

lldb/source/Plugins/ABI/MOS/MOSImaginaryRegisters.cpp
```diff
@@ -0,0 +1,239 @@
...
+  for (const auto &reg : regs) {
+    if (reg.byte_offset != LLDB_INVALID_INDEX32)
+      next_offset = std::max(next_offset, reg.byte_offset + reg.byte_size);
+    if (reg.regnum_remote != LLDB_INVALID_REGNUM)
+      next_regnum = std::max(next_regnum, reg.regnum_remote + 1);
+  }
+
+  // Add RC registers (8-bit)
+  for (uint32_t i = 0; i < num_rc_regs_; ++i) {
+    uint32_t dwarf_num = rc_base_dwarf_ + i;
```
mysterymath: This assumes the refactoring of `MOSRegisterInfo.td` done in #601. Since both PRs depend on this, it should be broken out so that we can submit it first.

lldb/source/Plugins/ABI/MOS/ABISysV_mos.cpp
```diff
@@ -0,0 +1,326 @@
...
+  UnwindPlan::Row row;
+
+  // CFA = normalized(S + 3) = ((S + 3) & 0xFF) | 0x0100
+  row.GetCFAValue().SetIsDWARFExpression(cfa_expr.data(), cfa_expr.size());
+
+  // PC = [normalized(S + 1)] - direct expression, NOT [CFA - 2]
+  // atDWARFExpression means: dereference the address computed by the
+  // expression.
+  UnwindPlan::Row::AbstractRegisterLocation pc_loc;
+  pc_loc.SetAtDWARFExpression(pc_expr.data(), pc_expr.size());
```
mysterymath: The value stored at S+1 by a JSR is not the return PC, but rather the return PC + 1.

lldb/source/Plugins/ABI/MOS/ABISysV_mos.cpp
```diff
@@ -0,0 +1,326 @@
...
+      "Setting return values not implemented for MOS");
+}
+
+ValueObjectSP ABISysV_mos::GetReturnValueObjectImpl(Thread &thread,
+                                                    CompilerType &type) const {
+  return ValueObjectSP();
+}
+
+bool ABISysV_mos::RegisterIsVolatile(const RegisterInfo *reg_info) {
+  // Nothing ever happens behind your back on MOS, so no volatile registers
```
mysterymath: `RegisterIsVolatile` actually means "call-clobbered": it is used to decide whether an unavailable caller register can be inherited from a younger frame. This information needs to be determined from the default calling convention.

lldb/source/Plugins/ABI/MOS/ABISysV_mos.cpp
```diff
@@ -0,0 +1,326 @@
...
+
+  plan_sp->AppendRow(std::move(row));
+  plan_sp->SetReturnAddressRegister(dwarf_pc);
+
+  return plan_sp;
+}
+
+UnwindPlanSP ABISysV_mos::CreateDefaultUnwindPlan() {
+  // The default unwind plan is used when we're in the middle of a function.
+  // For 6502, the return address is still on the hardware stack at S+1/S+2.
```
mysterymath: This isn't true; the compiler freely inserts `PHA` and `PLA` to save and restore values temporarily. In between those pairs, the function entry unwind plan will produce arbitrary garbage. Still, this is intended to be heuristic in the case that CFI or is unavailable, so using the function plan is fine; it just needs to be documented as a heuristic.

lldb/source/Plugins/ABI/MOS/ABISysV_mos.cpp
```diff
@@ -0,0 +1,326 @@
...
+  auto plan_sp = std::make_shared<UnwindPlan>(eRegisterKindDWARF);
+  plan_sp->SetSourceName("mos function-entry unwind plan");
+  plan_sp->SetSourcedFromCompiler(eLazyBoolNo);
+  plan_sp->SetUnwindPlanValidAtAllInstructions(eLazyBoolNo);
+  plan_sp->SetUnwindPlanForSignalTrap(eLazyBoolNo);
+
+  UnwindPlan::Row row;
+
+  // CFA = normalized(S + 3) = ((S + 3) & 0xFF) | 0x0100
+  row.GetCFAValue().SetIsDWARFExpression(cfa_expr.data(), cfa_expr.size());
```
mysterymath: The first row must issue `row.SetUnspecifiedRegistersAreUndefined(true)`. LLDB asserts on this in `GetFullUnwindPlanForFrame()`.

lldb/source/Plugins/ABI/MOS/ABISysV_mos.cpp
```diff
@@ -0,0 +1,326 @@
...
+             "register context");
+    return nullptr;
+  }
+
+  MOSImaginaryRegisters &imag_regs =
+      MOSImaginaryRegisters::GetOrCreate(*main_module, *m_mc_register_info_up);
+
+  return std::make_shared<MOSRegisterContext>(
+      *gdb_thread, concrete_frame_idx, reg_info_sp,
+      /*read_all_registers_at_once=*/false,
```
mysterymath: If we don't support this, register checkpointing will break. Accordingly, we need to override the two`ReadAllRegisterValues` and `WriteAllRegisterValues` variants (Buffer and Checkpoint) in RegisterContext to mark this as unavailable.

lldb/source/Plugins/ABI/MOS/ABISysV_mos.cpp
```diff
@@ -0,0 +1,326 @@
...
+//------------------------------------------------------------------
+static int GetDwarfRegNumByName(llvm::MCRegisterInfo *mc_info,
+                                llvm::StringRef name) {
+  if (!mc_info)
+    return -1;
+
+  // MCRegisterInfo stores names in uppercase
+  std::string mc_name = name.upper();
+
+  for (unsigned reg = 0; reg < mc_info->getNumRegs(); ++reg) {
```
mysterymath: nit: no braces, and break out ending var in the I/E style.

lldb/source/Plugins/ABI/MOS/ABISysV_mos.cpp
```diff
@@ -0,0 +1,326 @@
...
+//------------------------------------------------------------------
+// Unwind Plans
+//------------------------------------------------------------------
+
+/// Build a DWARF expression that computes a normalized hardware stack address.
+/// The 6502 hardware stack is at 0x0100-0x01FF. The S register is physically
+/// 8-bit, but some debuggers report it as 16-bit (e.g., MAME reports 0x01FE).
+/// This expression normalizes to always produce addresses in 0x0100-0x01FF:
+///   result = ((S + Offset) & 0xFF) | 0x0100
+static void buildNormalizedHardwareStackExpr(std::vector<uint8_t> &expr,
```
mysterymath: If `dwarf_s` is always set to the same value, maybe this should be a member function so the argument can be elided.

lldb/source/Plugins/ABI/MOS/ABISysV_mos.cpp
```diff
@@ -0,0 +1,326 @@
...
+}
+
+UnwindPlanSP ABISysV_mos::CreateFunctionEntryUnwindPlan() {
+  // Get the DWARF register numbers by name via MCRegisterInfo.
+  // These are looked up at runtime - no hardcoded constants.
+  // The actual DWARF numbers come from MOSRegisterInfo.td.
+  if (!m_mc_register_info_up)
+    return nullptr;
+
+  int dwarf_s = GetDwarfRegNumByName(m_mc_register_info_up.get(), "S");
```
mysterymath: `GetEHAndDwarfNums`

lldb/include/lldb/Target/ABI.h
```diff
@@ -179,6 +179,22 @@ class ABI : public PluginInterface {
     return std::nullopt;
   }
 
+  virtual lldb::RegisterContextSP
```
mysterymath: This only supports GDBRemoteRegisterContext and ThreadGDBRemote, so it should declare as such. And its name should probably reflect this.

lldb/source/Plugins/ABI/MOS/ABISysV_mos.cpp
```diff
@@ -0,0 +1,326 @@
...
+  // to avoid subtraction underflow when S wraps within the stack page.
+  //
+  // The expression computes: ((S + offset) & 0xFF) | 0x0100
+  // This ensures the result is always in the hardware stack range
+  // 0x0100-0x01FF.
+
+  // IMPORTANT: These expression vectors must be static because UnwindPlan
+  // stores raw pointers to the data, not copies. They must persist for the
+  // lifetime of the UnwindPlan.
+  static std::vector<uint8_t> cfa_expr;
```
mysterymath: For thread safety, this should be a static const struct initialized by immediately calling a lambda.

lldb/source/Plugins/ABI/MOS/ABISysV_mos.cpp
```diff
@@ -0,0 +1,326 @@
...
+  if (!initialized || cached_dwarf_s != dwarf_s) {
+    cfa_expr.clear();
+    pc_expr.clear();
+    s_expr.clear();
+
+    buildNormalizedHardwareStackExpr(cfa_expr, dwarf_s, 3); // CFA = S + 3
+    buildNormalizedHardwareStackExpr(pc_expr, dwarf_s, 1);  // PC at S + 1
+    buildNormalizedHardwareStackExpr(s_expr, dwarf_s, 2);   // S_prev = S + 2
+
+    cached_dwarf_s = dwarf_s;
```
mysterymath: `dwarf_s` isn't going to change; this should probably be an assertion.

lldb/source/Plugins/ABI/MOS/ABISysV_mos.h
```diff
@@ -0,0 +1,85 @@
...
+  bool GetArgumentValues(lldb_private::Thread &thread,
+                         lldb_private::ValueList &values) const override;
+
+  lldb_private::Status
+  SetReturnValueObject(lldb::StackFrameSP &frame_sp,
+                       lldb::ValueObjectSP &new_value) override;
+
+  bool CallFrameAddressIsValid(lldb::addr_t cfa) override {
+    // MOS uses 16-bit addresses, soft stack can be anywhere
+    return cfa <= 0xFFFF;
```
mysterymath: The ABI treats the call frame as the hardware call frame; this treats the soft stack as the call frame? Which is it? At the very least, this bears some discussion in a comment here.

lldb/source/Plugins/ABI/MOS/MOSImaginaryRegisters.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+// Static cache
+llvm::StringMap<MOSImaginaryRegisters> MOSImaginaryRegisters::s_cache_;
+
+//------------------------------------------------------------------
+// Helper to look up DWARF register number by name via MCRegisterInfo.
+// This is DRY - the actual DWARF numbers come from MOSRegisterInfo.td
+// through the MCRegisterInfo generated at LLVM build time.
+//------------------------------------------------------------------
+static int GetDwarfRegNumByName(llvm::MCRegisterInfo &mc_info,
```
mysterymath: Rather than duplicating this functionality with the `GetEHAndDwarfNums` in the ABI, we could precompute these DWARF nums in a table in the factory using access to the ABI object?

lldb/source/Plugins/ABI/MOS/MOSImaginaryRegisters.cpp
```diff
@@ -0,0 +1,239 @@
...
+  }
+  return -1;
+}
+
+MOSImaginaryRegisters &
+MOSImaginaryRegisters::GetOrCreate(Module &module,
+                                   llvm::MCRegisterInfo &mc_info) {
+  // Use UUID as cache key, fall back to file path
+  std::string key = module.GetUUID().GetAsString();
+  if (key.empty()) {
```
mysterymath: nit: no braces, here and throughout, for single statement control flow

PR #575, opened 2026-07-22
[compiler-rt] hand-tuned single-precision softfloat for MOS 6502

johnwbyrd:
## Summary

Adds MOS 6502 hand-tuned implementations of the four single-precision
IEEE 754 soft-float compiler-rt builtins: `__addsf3`, `__subsf3`,
`__mulsf3`, `__divsf3`.  Each is a naked C function containing one big
`asm volatile` block; all state lives in caller-saved zero-page slots
(`__rc2..__rc19`), so there is no callee-save tax and no stack frame.

## Numbers (vs upstream generic, on the existing compiler-rt Unit tests, under mos-sim)

| routine   | vectors    | isolated .text gen -> mos | full-test cycles gen -> mos |
|-----------|------------|---------------------------|------------------------------|
| `__addsf3`  | 25         | 3130 -> **675 B** (-78%)  | 33,774 -> **17,955** (-47%)     |
| `__subsf3`  | 22 (hand)  | -> **11 B** (trampoline)  | -> 19,429                       |
| `__mulsf3`  | 350+       | 2424 -> **681 B** (-72%)  | 2,310,376 -> **967,608** (-58%) |
| `__divsf3`  | 351        | 3305 -> **806 B** (-76%)  | 5,438,847 -> **535,174** (-90%) |

All results bit-exact on every upstream `Unit/*_test.c` vector (including
subnormals, subnormal-boundary rounding, exact-half tie-to-even,
NaN quieting, Inf/0 special cases).  The 10x `__divsf3` speedup is
because the generic path is Newton-Raphson (three iterations, each
needing a 32x32 multiply -- catastrophic on the 6502).  Shift-subtract
long division is a much better algorithmic fit.

`__subsf3` has no `Unit/subsf3_test.c` in compiler-rt; verified against
a 22-vector hand-rolled smoke test covering the IEEE 754 cases where
`a + (-b)` differs subtly from `a - b` (signed zero, NaN sign propagation,
subnormal boundaries).

## End-to-end validation

A 60x25 ASCII Mandelbrot demo (exercises all four routines) produces
**bit-identical output** vs generic softfloat (MD5 verified) at
**3.2x speed** (200M -> 638M cycles under `mos-sim`).

## Design notes

- One naked top-level function per file, one big `asm volatile` block
  inside.  All state in caller-saved zero page (`__rc2..__rc19`).  No
  callee-saved slots touched; no stack frame; LTO can still see the
  C-level signature for ABI/DCE analysis.
- Local labels use a per-function `.L__addsf3_*` / `.L__mulsf3_*` etc.
  prefix.  The `.L` keeps them out of the object symbol table; the
  function-name prefix keeps them unique across the LTO unit.  Without
  the per-function prefix, `.Lpack_a` in two sibling files collides
  during LTO because `.L` scoping is per-assembler-invocation, not
  per-source-file, and the mos-target build codegens the whole link
  into a single `.s`.
- Far branches (>127 bytes) use invert-and-jmp or a shared JMP
  trampoline to work around the LLVM MC assembler's silent 8-bit
  branch-offset truncation.

## Commits

1. `[MOS][compiler-rt] hand-tuned __addsf3 (naked asm, IEEE 754 binary32 add)`
2. `[MOS][compiler-rt] hand-tuned __mulsf3 (naked asm, IEEE 754 binary32 mul)`
3. `[MOS][compiler-rt] add __subsf3 as naked-asm trampoline into __addsf3`
4. `[MOS][compiler-rt] hand-tuned __divsf3 (naked asm, shift-subtract long division)`

## Test plan

- [x] `Unit/addsf3_test.c` (25 vectors) passes bit-exact under `mos-sim`
- [x] `Unit/mulsf3_test.c` (350+ vectors) passes bit-exact under `mos-sim`
- [x] `Unit/divsf3_test.c` (351 vectors) passes bit-exact under `mos-sim`
- [x] Hand-written 22-vector `__subsf3` smoke test passes bit-exact
- [x] `ninja runtimes/install-builtins-mos-unknown-unknown` builds and installs
- [x] Downstream `llvm-mos-sdk` libcrt.a rebuilds cleanly
- [x] End-to-end Mandelbrot via installed archive: bit-identical to generic, 3.2x speed

## Note on AI contribution

Substantial portions of this code contribution were written by Anthropic Claude Opus 4.8.  I have reviewed this code contribution by hand, and believe the quality of the generated code to be equivalent to that which I might have created without this tool.  Any errors or omissions in this code are my responsibility, not that of Claude Opus 4.8.

### DIFF
--- a/compiler-rt/lib/builtins/mos/addsf3.c
+++ b/compiler-rt/lib/builtins/mos/addsf3.c
@@ -1,0 +1,867 @@
+//===-- lib/mos/addsf3.c - Single-precision addition (MOS) -------*- C -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// __addsf3(a, b) -- IEEE 754 binary32 addition for MOS (6502).
+//
+// Rounding: round-to-nearest-ties-to-even, hardcoded.  There is no
+// support for fenv (no dynamic rounding mode, no inexact flag).  This
+// is a deliberate choice: MOS programs almost never manipulate the
+// floating-point environment, and the generic compiler-rt fenv machinery
+// (__fe_getround, __fe_raise_inexact) is ~30x more expensive per call
+// than the arithmetic it wraps.
+//
+// This file is intended to serve as the template for future single- and
+// double-precision soft-float routines on MOS (subsf3, mulsf3, divsf3,
+// adddf3, muldf3, ...).  The extensive reference material below is here
+// to make those follow-on implementations straightforward.
+//
+// ============================================================================
+// REFERENCE: IEEE 754 binary32 memory layout (little-endian)
+// ============================================================================
+//
+//   byte 3  : [S | EEEEEEE ]     S     = sign bit
+//   byte 2  : [E | MMMMMMM ]     EEEEE = 8-bit biased exponent, split
+//   byte 1  : [MMMMMMMM    ]     M     = 23-bit stored mantissa
+//   byte 0  : [MMMMMMMM    ]     (implicit leading 1 for normals)
+//
+// So the 8-bit biased exponent EEEEEEEE spans byte3's low 7 bits AND
+// byte2's high bit.  To extract it we shift bit 7 of byte2 into the carry
+// with ASL, then load byte3 and ROL through carry:
+//
+//     LDA byte2 ; ASL A    -- C = byte2[7] = E0 (low bit of exp)
+//     LDA byte3 ; ROL A    -- A = (byte3 << 1) | C = biased exp
+//
+// Number semantics by exponent field:
+//   exp == 0xFF, mantissa != 0 -- NaN (quiet if bit 22 (m3 bit 6) is set)
+//   exp == 0xFF, mantissa == 0 -- +/- infinity
+//   exp == 0x00, mantissa != 0 -- subnormal (value = M * 2^-149)
+//   exp == 0x00, mantissa == 0 -- +/- zero (sign preserved)
+//   exp in [1,254]             -- normal (value = 1.M * 2^(exp-127))
+//
+// For arithmetic we restore the implicit leading 1 for normal operands
+// so the mantissa is 24 bits [1.MMMMMMMMMMMMMMMMMMMMMMM].  We store it
+// in the top three bytes of a 4-byte tuple [m3:m2:m1:m0] where m0 is a
+// sub-LSB byte holding round + guard bits shifted down during alignment.
+// A separate `stk` byte accumulates a running sticky OR of bits that
+// fall off the bottom of m0.
+//
+//   4-byte working mantissa: [m3 : m2 : m1 : m0]  (little-endian; m0 low)
+//        m3 bit 7 = implicit leading 1 for normals (set on unpack)
+//        m1..m3   = 24-bit true mantissa
+//        m0       = sub-LSB byte (round + guard + partial sticky)
+//   +  1-byte stk = sticky OR of bits shifted below m0's bit 0
+//
+// ============================================================================
+// REFERENCE: MOS calling convention for this function
+// ============================================================================
+//
+// float __addsf3(float a, float b) with the standard llvm-mos ABI:
+//
+//   INPUT:  a's bytes in A, X, __rc2, __rc3  (byte 0 in A ... byte 3 in rc3)
+//           b's bytes in __rc4, __rc5, __rc6, __rc7
+//   OUTPUT: result bytes in A, X, __rc2, __rc3
+//
+//   Caller-saved:  A, X, Y, flags, __rc2..__rc19 (RS1..RS9)
+//   Callee-saved:  __rc0/__rc1 (RS0 = soft SP), __rc20..__rc31 (RS10..RS15)
+//
+// Because this is `__attribute__((naked, noinline))`, the compiler emits
+// no prologue or epilogue.  We are responsible for everything, but the
+// benefit is enormous: since we confine all state to caller-saved slots
+// we pay zero cycles/bytes to save/restore callee-saved registers.
+//
+// ============================================================================
+// REFERENCE: zero-page layout during arithmetic
+// ============================================================================
+//
+// Fields are chosen so that on entry the incoming byte order (A, X in
+// registers; b's bytes starting at __rc4) maps cleanly to our working
+// slots with the fewest moves during unpack.
+//
+//     Slot     Name    Purpose
+//     ------   ------  ---------------------------------------------------
+//     __rc2    B_M0    b's sub-LSB byte (0 on entry, holds R/G bits)
+//     __rc3    B_M1    b's low mantissa byte  (was b_byte0 on entry)
+//     __rc4    B_M2    b's mid mantissa byte  (was b_byte1 on entry)
+//     __rc5    B_M3    b's high mantissa byte (implicit 1 in bit 7 for normal)
+//     __rc6    B_EXP   b's biased exponent
+//     __rc7    B_SIGN  b's sign in bit 7 (0x00 or 0x80)
+//     __rc8    B_STK   b's sticky bit (0 or nonzero)
+//
+//     __rc9    TMP     scratch used in align/pack/round
+//
+//     __rc10   A_M0    a's sub-LSB byte
+//     __rc11   A_M1    a's low mantissa byte
+//     __rc12   A_M2    a's mid mantissa byte
+//     __rc13   A_M3    a's high mantissa byte
+//     __rc14   A_EXP   a's biased exponent
+//     __rc15   A_SIGN  a's sign in bit 7
+//     __rc16   A_STK   a's sticky bit
+//
+//     __rc17   TMP2    scratch used in align
+//     __rc18/19 unused
+//
+// After the magnitude-compare swap, `a` always holds the operand with
+// the larger absolute value, so the special-case dispatch only needs to
+// classify a (and check b in a handful of sub-cases).
+//
+// ============================================================================
+// REFERENCE: MOS assembler gotchas we work around
+// ============================================================================
+//
+// 1. Assembler-local labels use the ".L" prefix (stripped from the object
+//    symbol table).  Under LTO the whole link is codegenned into one .s,
+//    so ".L" scoping is per-assembler-invocation, NOT per-source-file --
+//    two mos/*.c files defining ".Lpack_a" WILL collide.  We work around
+//    this with a per-function prefix (".L__addsf3_pack_a" etc.) so every
+//    label is unique across the LTO unit.
+//    Numeric labels (1:, 2:, ..., 32:) work the same way inside inline
+//    asm: they become .L__addsf3_tmpNN and each `Nf` refers to the next `N:`
+//    forward.  Do NOT reuse the same numeric label multiple times.
+//
+// 2. Out-of-range branch offsets are silently TRUNCATED to 8 bits rather
+//    than reported as an error.  A BEQ/BNE/BCC/BCS whose target is more
+//    than +127/-128 bytes away will wrap around to a totally different
+//    address and execute garbage.  For any conditional branch that
+//    reaches a label past that limit we use the invert-and-jump pattern:
+//        bne .L__addsf3_skip / jmp .L__addsf3_far_target / .L__addsf3_skip:
+//    OR, when several far branches share a common target, we place a
+//    small trampoline (jmp .L__addsf3_far_target) within byte range of all of
+//    them and branch to the trampoline.  See .L__addsf3_tramp_pack_a below.
+//
+// 3. Even-numbered __rcN slots (rc2, rc4, ...) are declared with
+//    PROVIDE() in the linker script, so a platform can reorder them.
+//    The only guaranteed-contiguous groups are pointer pairs:
+//    (rc0,rc1), (rc2,rc3), (rc4,rc5), ...  This means we can NOT use
+//    zp,x indexed addressing across our named slots -- every access has
+//    to be by name.  See the do_swap section below.
+//
+// ============================================================================
+// REFERENCE: algorithm phases
+// ============================================================================
+//
+//   1. UNPACK A: split A/X/rc2/rc3 into sign/exp/mantissa/sticky in
+//      __rc10..__rc16.  Restore implicit-1 bit in m3 for normals.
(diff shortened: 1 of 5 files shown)


compiler-rt/lib/builtins/mos/addsf3.c
```diff
@@ -0,0 +1,867 @@
...
+        "lda " A_M0 "\n" "sbc " B_M0 "\n" "sta " A_M0 "\n"
+        "lda " A_M1 "\n" "sbc " B_M1 "\n" "sta " A_M1 "\n"
+        "lda " A_M2 "\n" "sbc " B_M2 "\n" "sta " A_M2 "\n"
+        "lda " A_M3 "\n" "sbc " B_M3 "\n" "sta " A_M3 "\n"
+        // Merge b's sticky into a's sticky.
+        "lda " B_STK "\n"
+        "beq 9f\n"
+        "lda #1\n"
+        "sta " A_STK "\n"
+        "9:\n"
```
wbniv, line 606: This sticky-bit merge is correct for the ADD path just above (dropped `b` bits → true sum is larger → bias round up), but it's backwards here: truncating `b`'s low bits during alignment means the computed *difference* is slightly too **large**, not too small, so a nonzero `B_STK` should bias the round *down* for subtraction, not up.

Suggested fix: ripple-decrement the `[A_M3:A_M2:A_M1:A_M0]` tuple by one unit and force sticky unconditionally — the true difference is `computed − ε` for `0 < ε < 1` unit, i.e. representable as `(computed − 1)` plus a strictly-positive inexact remainder.

This can never underflow an all-zero tuple: whenever `B_STK` can be nonzero, an alignment shift occurred, and a logical right-shift always zero-fills its vacated top bit — so `B_M3`'s bit 7 is provably 0 post-shift while `A_M3`'s bit 7 is provably 1 here (`exp != 0`), so the raw subtraction result can never be exactly zero when `B_STK` is set.

Verified: 1M-pair model re-fuzz → 0 mismatches; replay of the exact seed that originally found 63/300,000 failures here → 0/300,000; compiles clean with a real MOS `clang` at every optimization level.

```suggestion
        // FIX: a nonzero B_STK here means b's true value is slightly
        // LARGER than the truncated value we just subtracted (its low
        // bits were dropped during alignment), so the difference we
        // just computed is slightly TOO LARGE -- by a strictly-positive
        // amount less than one unit of A_M0's LSB.  This is the
        // opposite correction from the same-sign ADD path above (where
        // a nonzero B_STK correctly biases the round UP): here it must
        // bias the round DOWN.  Implement that by borrowing one full
        // unit out of the [A_M3:A_M2:A_M1:A_M0] tuple (ripple decrement)
        // and then marking sticky unconditionally -- the leftover
        // fractional amount (strictly between 0 and 1 unit) is never
        // exactly representable, so any later round decision must treat
        // this position as inexact. (The ripple decrement can never
        // underflow an all-zero tuple: whenever B_STK can be nonzero, an
        // alignment shift occurred, and a logical right-shift always
        // zero-fills its vacated top bit, so B_M3's bit 7 is provably 0
        // post-shift while A_M3's bit 7 is provably 1 here [exp != 0],
        // so the tuple this subtract produces can never be exactly zero
        // when B_STK is set.)
        "lda " B_STK "\n"
        "beq .L__addsf3_sub_stk_done\n"
        "dec " A_M0 "\n"
        "lda " A_M0 "\n"
        "cmp #$ff\n"
        "bne .L__addsf3_sub_stk_set\n"
        "dec " A_M1 "\n"
        "lda " A_M1 "\n"
        "cmp #$ff\n"
        "bne .L__addsf3_sub_stk_set\n"
        "dec " A_M2 "\n"
        "lda " A_M2 "\n"
        "cmp #$ff\n"
        "bne .L__addsf3_sub_stk_set\n"
        "dec " A_M3 "\n"
        ".L__addsf3_sub_stk_set:\n"
        "lda #1\n"
        "sta " A_STK "\n"
        ".L__addsf3_sub_stk_done:\n"
```

compiler-rt/lib/builtins/mos/addsf3.c
```diff
@@ -0,0 +1,867 @@
...
+        "rol " A_M0 "\n"                     // shift with sticky into new m0[0]
+        "rol " A_M1 "\n"
+        "rol " A_M2 "\n"
+        "rol " A_M3 "\n"
+        "lda #0\n"
+        "sta " A_STK "\n"
+        "dec " A_EXP "\n"
+        "jmp .L__addsf3_norm_bit_loop\n"
+
+        ".L__addsf3_norm_done:\n"
```
wbniv, line 693: This is the severe bug: `exp==1` is the transition into subnormal representation, not one more normal-to-normal renormalization step. IEEE-754 subnormals (biased exp field 0) reuse the **same** reference scale as the smallest normal (biased exp field 1) — they just drop the implicit leading 1 — so at `exp==1` with bit 7 still clear, the current (unshifted) mantissa is already the correct subnormal encoding. Shifting-and-decrementing across that boundary (what the loop does today, because the `beq` tests the pre-decrement value) silently doubles the result — 31% of nonzero-subnormal results in a targeted fuzz came back exactly 2×.

The `NORMALIZE OVERSHOOT CORRECTION` block below handles the mirror-image boundary (bit 7 becoming set exactly as exp reaches 0) but is one-directional and doesn't cover this case.

This suggestion adds the `exp==1` early-exit plus its companion branch target in one block (the intervening shift body is unchanged).

Verified: replay of the seed that originally found 8,267/26,594 doubled subnormal results → 0/26,594; 1M-pair re-fuzz clean; compiles at every optimization level (+37 B total with the other fix — 712 B vs 675 B, still 77% below the generic `addsf3.c`).

```suggestion
        "beq .L__addsf3_norm_done\n"                  // subnormal -- stop
        // FIX: exp==1 is the transition into subnormal representation,
        // not one more normal-to-normal renormalization step.  Normal
        // exponents E and E-1 (both >=1) share a "shift mantissa left 1,
        // decrement exp" invariant that preserves value; that invariant
        // does NOT extend to E=1 -> E=0, because IEEE 754 subnormals
        // (biased exp field 0) reuse the SAME reference scale as the
        // smallest normal (biased exp field 1) -- they just drop the
        // implicit leading 1.  So at exp==1 with bit 7 still clear, the
        // CURRENT (unshifted) mantissa is already the correct subnormal
        // encoding: relabel exp to 0 and stop, without shifting.
        // Shifting here as if crossing a normal exponent boundary
        // silently doubles the result. (A same-shape check already
        // exists a few lines below, in the OVERSHOOT CORRECTION block,
        // for the opposite boundary condition -- bit 7 becoming set
        // exactly as exp reaches 0 -- but that fix is one-directional
        // and does not cover this case.)
        "cmp #1\n"
        "beq .L__addsf3_norm_to_subnormal\n"
        "lda " A_STK "\n"
        "lsr a\n"                            // C <- stk bit 0
        "rol " A_M0 "\n"                     // shift with sticky into new m0[0]
        "rol " A_M1 "\n"
        "rol " A_M2 "\n"
        "rol " A_M3 "\n"
        "lda #0\n"
        "sta " A_STK "\n"
        "dec " A_EXP "\n"
        "jmp .L__addsf3_norm_bit_loop\n"

        ".L__addsf3_norm_to_subnormal:\n"
        "lda #0\n"
        "sta " A_EXP "\n"
        "jmp .L__addsf3_norm_done\n"

        ".L__addsf3_norm_done:\n"
```

wbniv (changes requested):
Reviewed the four new routines by hand-tracing the 6502 assembly, transcribing each into an independent bit-exact Python model, and fuzzing against `numpy` float32 arithmetic as ground truth (random pairs + targeted edge classes: subnormals, over/underflow boundaries, signed zero, NaN/Inf). Divergences were traced back to the exact instructions to confirm root cause.

**`mulsf3.c`**: no bugs found (500k random + ~371k targeted subnormal/boundary cases, 0 mismatches).

**`divsf3.c`**: no correctness bugs (300k random + exhaustive boundary scan). Two non-blocking notes for a follow-up: the post-loop "R≥D cleanup" step looks like unreachable dead code — the loop invariant `R<D` holds at every iteration regardless of which branch is taken, so it never fires (the commit message's stated rationale for it doesn't hold given that invariant); and the underflow shift-count computation drops the high byte of a 16-bit signed exponent, which is safe today only because of an exponent-range bound that isn't documented near the code.

**`addsf3.c`/`__subsf3`** (the latter tail-calls the former, inheriting both bugs below): two real bugs, one severe.

1. **Severe — results come back exactly 2× too large for `a - b` whenever the correctly-rounded difference is a subnormal float** (< ~1.175e-38), i.e. subtracting two nearly-equal small floats. In a targeted fuzz of same-exponent opposite-sign subtractions, 31% of the cases whose true result is a nonzero subnormal came back doubled. Root cause: the normalize bit-loop tests `A_EXP == 0` *before* deciding whether to shift, using the pre-decrement value — when `A_EXP` is 1 (always true right before crossing into subnormal territory) that check is false, so the loop performs one more shift-and-decrement than it should. IEEE-754 subnormals share the *same* reference scale as the smallest normal rather than getting an additional halving, so the loop needs to stop one iteration earlier at that boundary. Inline suggestion below covers the check and its companion branch target in one block.

2. **Alignment sticky bit gets merged with the wrong sign on the subtract path.** When `b`'s low bits are dropped during alignment (`B_STK` set), the code biases the round decision *up*, identically to the add path above it. Correct for addition, backwards for subtraction: truncating `b` down makes the computed difference slightly too *large*, so a nonzero alignment sticky should bias the round *down* here. ~63/300k mismatches in random fuzzing, each exactly +1 ULP — the classic sticky-bit sign flip in hand-written FPU subtract logic. Suggestion below.

Both are hit by ordinary inputs, not adversarial construction — plain `random.getrandbits` finds the severe one at a 31% in-class rate. That tracks with the PR's test coverage: `addsf3_test.c`'s 25 fixed vectors don't happen to probe this magnitude relationship. Related: the 22-vector `__subsf3` smoke test described in the PR text isn't in this diff — the PR touches exactly 5 files (`CMakeLists.txt` + the 4 new `.c` files) — so that claim isn't currently verifiable by CI or reviewers, and `subsf3.c` (which inherits both bugs) would ship with no in-tree regression coverage.

I verified both fixes before suggesting them: updated the model and re-fuzzed 1,000,000 fresh random pairs (0 mismatches), replayed the exact seeds that originally found the failures (0/26,594 and 0/300,000), and compiled the patched file with a real MOS `clang` at every optimization level — clean, +37 B (712 B vs 675 B; still a 77% reduction over the generic `addsf3.c`). I did not get to an actual `mos-sim` execution run (sysroot layout friction on my end) — flagging that honestly rather than claiming it.

Nice work overall — the algorithmic choices (especially shift-subtract division over Newton-Raphson for `__divsf3`) are well-reasoned and the size/cycle numbers are real. Given the severity of bug #1 I'd hold this from merging until it's fixed; happy to help verify further if useful.

johnwbyrd:
wbniv, some really good work here. I wonder if I can convince you to get some of the failing test cases into compiler-rt/test/builtins/Unit/*_test.c?  This would presumably fail my code, while passing your changes on all platforms.  Since you can see your Python code, you might be able to trivially add those fuzzed cases into the unit tests.  Thanks!

compiler-rt/lib/builtins/mos/addsf3.c
johnwbyrd: All the FIX:/CURRENT: business smells of AI and has no place here.  The code should be rewritten correctly, and source control should track the changes, not vice versa.

compiler-rt/lib/builtins/mos/addsf3.c
```diff
@@ -0,0 +1,867 @@
...
+#define A_M3   "__rc13"
+#define A_EXP  "__rc14"
+#define A_SIGN "__rc15"
+#define A_STK  "__rc16"
+// __rc17..__rc18 are free; align scratch reuses __rc9 (TMP).
+#define TMP    "__rc9"
+#define TMP2   "__rc17"                      // second scratch (align)
+
+__attribute__((naked, noinline))
+COMPILER_RT_ABI float __addsf3(float af, float bf) {
```
mysterymath, line 193: As soon as I see naked noinline into an asm volatile, my thought is, "This should have just been a .S file". Was there a compelling reason not to use a .S file here? If not, well, this should have just been a .S file; that would tend to make this dramatically easier to read.

johnwbyrd: You were right, and my rationale for naked-C was wrong.

I'd been carrying an assumption that LTO was what did dead-code elimination on unused routines — the theory being that a `naked` C function preserved a C-level signature the LTO optimizer could reason about, whereas a `.S` file was opaque to it, so `.S` would defeat DCE and bloat the binary. That's not how any of this works on the mos toolchain.

Quick bench: minimal `.S` with a 128-byte `__unused_bloat` symbol that `main` never references. Under `mos-sim-clang -Os` (LTO on by default) the bytes are stripped — but they're *also* stripped with `-fno-lto`. What actually brings them back is `-Wl,--no-gc-sections`. Grepping the driver's `-v` output, the mos toolchain hard-codes `-ffunction-sections -fdata-sections` at compile time and `--gc-sections` at link time, and archive-linker semantics won't pull an unreferenced `.o` from a `.a` in the first place. So `.S` in a static archive is triple-DCE'd via mechanisms that predate LTO entirely.

And even if the LTO story had been right, a naked function whose body is `asm volatile` is opaque to the LTO optimizer anyway — clang can't inline into it, inspect it, or reason about its clobbers — so the thing LTO offers over `.S` is essentially zero.

I'll rewrite the four routines as `.S`. That also makes the label-scoping issue go away (each `.s` becomes its own assembler invocation instead of everything getting concatenated into one LTO codegen output) and drops the naked/asm-volatile awkwardness. Thanks for the pushback.

johnwbyrd:
Closing this — will open a fresh PR once the rewrite lands.

Summary of what came out of this review:

- @wbniv found two real subtract-path bugs in `__addsf3` (severe subnormal-result doubling; alignment sticky-bit sign flip) that inherit into `__subsf3`. Reproduced both against the current tree.
- @mysterymath was right that naked-C + `asm volatile` was the wrong shape. I'd been carrying a wrong assumption that LTO was needed for DCE of unused routines; the mos toolchain does DCE via `--gc-sections` + archive semantics regardless of LTO, and a naked+asm-volatile function is opaque to LTO's optimizer anyway.

Plan for the follow-up PR: rewrite all four routines as `.S` files with the two bugs fixed. I've already got fuzz-derived regression coverage staged locally (extended `addsf3_test.c` plus a new `subsf3_test.c`); those will ship in the new PR so CI catches these bug shapes for anyone reimplementing the softfloat builtins on any target.

Thanks to both of you for the careful reads.

johnwbyrd:
Please see https://github.com/llvm-mos/llvm-mos/pull/597 for continuation.

PR #577, opened 2026-07-26
[MOS] Legalize G_SCMP/G_UCMP via lowerThreewayCompare

### LINKED ISSUE #576: Backend abort: unable to legalize G_SCMP/G_UCMP (C three-way compare)
The standard C three-way-compare ("spaceship") idiom crashes the backend at every integer width, on
every `-O` level, with and without LTO:

```c
// crash.c — any qsort comparator written this way triggers it
int cmp(short a, short b) { return (a > b) - (a < b); }
```

```
$ mos-clang --target=mos -mcpu=mosw65816 -Os -c crash.c
fatal error: error in backend: unable to legalize instruction:
  %2:_(s16) = G_SCMP %0:_, %1:_ (in function: cmp)
```

Clang/instcombine canonicalize `(a > b) - (a < b)` to the generic `llvm.scmp`/`llvm.ucmp`
intrinsics (introduced LLVM 19), which reach GlobalISel as `G_SCMP`/`G_UCMP` — and
`MOSLegalizerInfo` has no rule for either opcode, so legalization aborts. Unsigned comparators
(`unsigned`/`unsigned long` operands) hit the same wall via `G_UCMP`.

Practical impact: any program that sorts with a spaceship comparator — the idiomatic `qsort`
callback — fails to build for MOS. We hit it via a SNES demo sorting with `qsort` at `-Os`
(both the `-fno-lto` object path and the LTO link path reproduce).

IR repro (same crash under `llc`, reproduced 2026-07-26 on a pristine build of `main` @ `8be054612` —
`LLVM ERROR: unable to legalize instruction: %2:_(s8) = G_SCMP %0:_(s8), %1:_ (in function: scmp_i8)`):

```llvm
define i16 @cmp(i16 %a, i16 %b) {
  %r = call i16 @llvm.scmp.i16.i16(i16 %a, i16 %b)
  ret i16 %r
}
```

Fix incoming as a PR: a one-line legalizer rule routing `{G_SCMP, G_UCMP}` to `.lower()`, which
expands through LLVM's own `LegalizerHelper::lowerThreewayCompare` into the icmp+select shapes the
backend already legalizes — no generic-LLVM change. Verified on a pristine checkout of current
`main` plus a `llc` lit test covering s8/s16 results with s8/s16/s32/s64 operands, signed and
unsigned; downstream we have additionally soak-tested the exact change inside a 65816 fork where
sorted-demo output is differentially verified against a host oracle on two emulators — playable
in-browser at [qsortviz](https://biohack.net/snes/qsortviz/) (the demo that found this),
[spaceship](https://biohack.net/snes/spaceship/), [ucmprank](https://biohack.net/snes/ucmprank/),
[trimerge](https://biohack.net/snes/trimerge/), [keycmp64](https://biohack.net/snes/keycmp64/).

wbniv:
`MOSLegalizerInfo` had no rule for `G_SCMP`/`G_UCMP`, so the C three-way-compare idiom
`(a > b) - (a < b)` — which clang/instcombine canonicalize to `llvm.scmp`/`llvm.ucmp` — aborted the
backend (`unable to legalize instruction: ... = G_SCMP ...`) at every integer width, every `-O`
level, with and without LTO. Any `qsort` spaceship comparator fails to build for MOS.

The fix is one line, placed next to the analogous min/max lowering:

```cpp
getActionDefinitionsBuilder({G_SCMP, G_UCMP}).lower();
```

routing both opcodes through LLVM's existing `LegalizerHelper::lowerThreewayCompare` (an
icmp+select expansion the backend already legalizes end to end). No generic-LLVM change, no new
pseudo, no effect on any currently-compiling code (the opcodes previously always aborted).

Includes `llvm/test/CodeGen/MOS/scmp-ucmp.ll`: `llc -verify-machineinstrs` over signed and unsigned
three-way compares with s8/s16/s32 results and s8/s16/s32/s64 operands — every case aborts without
the fix and compiles verifier-clean with it. Each function also pins that the expansion stays
inline (`CHECK-NOT: jsr` between its label and its `rts`): no width falls back to a libcall.

Testing beyond the lit test: the identical change has been soak-tested in our 65816 development
fork, where five qsort-based SNES demos are differentially verified — host-computed result ==
on-console result on two independent emulators (MAME + bsnes-jg), `-verify-machineinstrs` clean —
and are **playable in-browser** (bsnes-jg WASM, each page's "Verify fidelity" button re-runs the
WRAM self-check live):

- [qsortviz](https://biohack.net/snes/qsortviz/) — the demo that found the crash (libc `qsort`,
  spaceship comparator)
- [spaceship](https://biohack.net/snes/spaceship/) — signed comparators at int8/16/32/64 keys
- [ucmprank](https://biohack.net/snes/ucmprank/) — the unsigned half (`G_UCMP` at u16/u32/u64)
- [trimerge](https://biohack.net/snes/trimerge/) — the scmp result consumed as control flow
- [keycmp64](https://biohack.net/snes/keycmp64/) — chained 64-bit tie-break (`G_SCMP s64` ×2/call)

Fixes #576.

### DIFF
--- a/llvm/lib/Target/MOS/MOSLegalizerInfo.cpp
+++ b/llvm/lib/Target/MOS/MOSLegalizerInfo.cpp
@@ -217,6 +217,13 @@ MOSLegalizerInfo::MOSLegalizerInfo(const MOSSubtarget &STI) {
 
   getActionDefinitionsBuilder({G_SMIN, G_SMAX, G_UMIN, G_UMAX}).lower();
 
+  // Three-way compares (llvm.scmp/ucmp — the C `(a>b)-(a<b)` spaceship idiom and libc-style
+  // comparators) have no native form on the 6502/65816; lower them to icmp+select primitives the
+  // backend already legalizes (LegalizerHelper::lowerThreewayCompare). Without this the backend
+  // aborts with "unable to legalize G_SCMP" on any qsort-style comparator. Not accum-specific —
+  // it fires at every width and in default/+mos-a16/+mos-xy16 alike.
+  getActionDefinitionsBuilder({G_SCMP, G_UCMP}).lower();
+
   getActionDefinitionsBuilder(G_ABS).custom();
 
   // Odd operations are handled via even ones: 6502 has only ADC/SBC.
--- a/llvm/test/CodeGen/MOS/scmp-ucmp.ll
+++ b/llvm/test/CodeGen/MOS/scmp-ucmp.ll
@@ -1,0 +1,71 @@
+; RUN: llc -verify-machineinstrs < %s | FileCheck %s
+;
+; The C three-way-compare ("spaceship") idiom `(a > b) - (a < b)` is
+; canonicalized by clang/instcombine to the generic llvm.scmp/llvm.ucmp
+; intrinsics (G_SCMP/G_UCMP in GlobalISel). The MOS legalizer had no rule for
+; them, so any program qsort-ing with such a comparator aborted with
+; "unable to legalize instruction: ... = G_SCMP ...". These now lower via
+; LegalizerHelper::lowerThreewayCompare (icmp + select expansion the backend
+; already legalizes). The checks pin each function's successful lowering to a
+; return and that the expansion stays inline (CHECK-NOT: jsr) — a regression
+; to either an abort or a libcall fails; the exact expansion is icmp/select
+; codegen owned elsewhere.
+
+target datalayout = "e-m:e-p:16:8-p1:8:8-i16:8-i32:8-i64:8-f32:8-f64:8-a:8-Fi8-n8"
+target triple = "mos"
+
+; CHECK-LABEL: scmp_i8:
+; CHECK-NOT: jsr
+; CHECK: rts
+define i8 @scmp_i8(i8 %a, i8 %b) {
+  %r = call i8 @llvm.scmp.i8.i8(i8 %a, i8 %b)
+  ret i8 %r
+}
+
+; CHECK-LABEL: ucmp_i8:
+; CHECK-NOT: jsr
+; CHECK: rts
+define i8 @ucmp_i8(i8 %a, i8 %b) {
+  %r = call i8 @llvm.ucmp.i8.i8(i8 %a, i8 %b)
+  ret i8 %r
+}
+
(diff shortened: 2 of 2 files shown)


wbniv:
Updated (`e54ef47` → `67020e2`, amended): the lit test now pins that the expansion stays inline — `CHECK-NOT: jsr` between each function's label and its `rts` (verified: no width emits a libcall, s64 operands included) — and adds an s32-result case (`scmp_i32_i16`). Fix unchanged.

llvm/lib/Target/MOS/MOSLegalizerInfo.cpp
```diff
@@ -217,6 +217,13 @@ MOSLegalizerInfo::MOSLegalizerInfo(const MOSSubtarget &STI) {
 
   getActionDefinitionsBuilder({G_SMIN, G_SMAX, G_UMIN, G_UMAX}).lower();
 
+  // Three-way compares (llvm.scmp/ucmp — the C `(a>b)-(a<b)` spaceship idiom and libc-style
```
mysterymath: The comment here is a superfluous; `.lower()` is essentially the default behavior for anything we don't decide to handle "specially". We would want a comment if we went `.custom()`, and even then it would describe the rationales for why we were lowering it that way.

llvm/test/CodeGen/MOS/scmp-ucmp.ll
```diff
@@ -0,0 +1,71 @@
+; RUN: llc -verify-machineinstrs < %s | FileCheck %s
+;
+; The C three-way-compare ("spaceship") idiom `(a > b) - (a < b)` is
```
mysterymath, line 3: Classic AI-ism; we don't need a comment saying what the state of the codebase was back when it was worse than it is now... especially when the way it was worse was "lol we missed that one". So only the first and last sentences here are useful.

mysterymath (approved):
LGTM after you address review comments and solve the merge conflict.

wbniv:
Addressed—thanks. I removed the redundant comment above `.lower()` and reduced the test preamble to only the intrinsic’s origin and the test’s intent. I also rebased onto current `main`, preserving the newer scalarization rules. The focused `llc -verify-machineinstrs | FileCheck` test passes.

PR #578, opened 2026-07-26
[MOS] Recompute loop liveness after copy forwarding

wbniv:
MOSCopyOpt can forward a loop-header `A = COPY Y` through reaching `Y = COPY A` definitions and remove the restore. This changes the register that must remain live around the back edge from Y to A.

A single post-order liveness update visits the latch before the header and can leave A absent from the latch's live-ins. Later CmpZero lowering then treats A as scratch and overwrites the loop-carried value. In the CRC reproducer this makes the next rotate consume the old accumulator value.

Propagate live-ins to a fixed point with `fullyRecomputeLiveIns` before dead-copy cleanup. The per-block live-in recompute inside the cleanup loop stays: it is what lets a copy in a predecessor become dead in the same pass after a successor's dead copy is erased. Keep the entry block's ABI live-ins intact. Remove the original shouldCoalesce workaround; ordinary rotate coalescing remains enabled.

The reduced MIR regression checks that A is live through the latch and that the copy-optimization/pseudo-expansion/late-optimization pipeline does not overwrite it. Both checks fail without the fix. A second test pins two- and three-block dead-copy chains so the per-block recompute cannot be dropped by accident. The three-block case also checks final assembly for the absence of leftover register transfers.

This replaces the original PR's diagnosis: register allocation emits the required restore correctly. The failure occurs later when copy forwarding changes the loop's liveness and the update does not propagate around the back edge. The defect is confined to the target-specific MOSCopyOpt pass; no generic LLVM change is involved.

Validation: MOS CodeGen suite at `b4749221bf37`: 80 pass, 1 unsupported. Both regression assertions fail without the fix. The reconstructed SNES demo changes from host/ROM CRC mismatch (0x7F81 / 0xC57C) to agreement at 0x7F81 in bsnes. Compile-time cost of the fixed-point recompute, measured as retired instructions over a 176-file corpus at -O2: +0.8% overall, +2.4% on the worst loop-dense kernel; on-CPU time unchanged.

### DIFF
--- a/llvm/lib/Target/MOS/MOSRegisterInfo.cpp
+++ b/llvm/lib/Target/MOS/MOSRegisterInfo.cpp
@@ -647,6 +647,21 @@ bool MOSRegisterInfo::shouldCoalesce(
     const TargetRegisterClass *NewRC, LiveIntervals &LIS) const {
   const auto &MRI = MI->getMF()->getRegInfo();
 
+  // Don't coalesce two shift/rotate-referenced values together into the A-only Ac
+  // class. The accumulator is the only register ASL/LSR/ROL/ROR can operate on, so
+  // an Ac value is pinned to A for its whole live range. When such a value is also
+  // loop-carried across an inner conditional whose other arm needs A (e.g. an
+  // inlined CRC16 bit loop: the bit-test reloads the pre-rotate byte into A, while
+  // the rotated value must survive to the next iteration), pinning it to A-only
+  // leaves the allocator no register to evict it to on the skip path — it strands
+  // the live value in Y while the back-edge rotate reads a stale A, miscompiling
+  // silently (the machine verifier does not catch it). Keeping the copy live lets
+  // the value occupy the broader AImag8/AY class and spill/transfer as needed.
+  if (NewRC == &MOS::AcRegClass &&
+      referencedByShiftRotate(MI->getOperand(0).getReg(), MRI) &&
+      referencedByShiftRotate(MI->getOperand(1).getReg(), MRI))
+    return false;
+
   // Don't coalesce Imag8 and AImag8 registers together when used by shifts or
   // rotates.  This may cause expensive ASL zp's to be used when ASL A would
   // have sufficed. It's better to do arithmetic in A and then copy it out.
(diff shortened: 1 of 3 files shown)


wbniv:
Updated (`1824492` → `edc9bbd`, amended): added `coalesce-rotate-ac-no-pessimize.ll`, pinning the cost side — common straight-line shift chains (`i8`/`i16` `<< 2`) keep their tight `asl`/`rol` form with the guard in place, byte-identical to the unguarded output. Also rewrote the PR body: a precise scope section (what the guard fixes vs. the downstream RA behavior it deliberately doesn't change), exact Csmith accounting, why no end-to-end miscompile test is extractable, and dropped the embedded patch listing that had drifted from the real diff. Fix unchanged.

mysterymath:
Yeah, reading the PR description, I reheheaaally don't like that this is papering over a bug elsewhere. That might *work*, but we really should get to the bottom of why this is happening. Especially since fixing upstream bugs is part of the game, *and* since I'm an upstream llvm contributor, and if we fix something wrong there, well, that's good for everyone.

wbniv:
You were right, and it wasn't the coalescer or RA at all. Got to the bottom of it; the branch and description are replaced (`edc9bbd` → `b474922`).

The coalescer's join and greedy's allocation were both correct — the rewriter emits the `A = COPY Y` restore at the loop header exactly as it should. What eats it is `MOSCopyOpt`: it forwards that copy away (legitimately — the value is already in Y), which changes the register that has to stay live around the back edge from Y to A. Then its liveness update is a *single* post-order pass over the blocks, which visits the latch before the header, so the header's new `$a` live-in never propagates to the latch. `mos-late-opt` later sees A dead at the latch and uses it as scratch when lowering the `CmpBrZero` — the stale-A read that looked like an RA bug from the outside. Pass-by-pass evidence in the description.

Fix is a `fullyRecomputeLiveIns` before the dead-copy cleanup so live-ins reach a fixed point first. The `shouldCoalesce` guard and both its tests are gone; ordinary rotate coalescing is back on. It's entirely inside the target pass — nothing upstream was wrong, so nothing to send upstream, sorry.

Regression is a reduced MIR loop: fails on both prefixes without the fix (latch lacks `$a`; the bad `LDImag8` reload appears), and the original CRC demo goes from a host/ROM mismatch to agreement. The recompute costs +0.8% retired instructions across a 176-file corpus, worst case +2.4% on a loop-dense kernel.

PR #579, opened 2026-07-26
[lld][MOS] Document the flat-output .elf companion

wbniv:
Documents the undocumented `<output>.elf` companion that `ld.lld` writes when a linker script uses the llvm-mos custom `OUTPUT_FORMAT { FULL/TRIM }` mechanism.

## What happens

For a flat-binary output such as an `.sfc`, `.nes`, or `.a26`, one link produces:

- `<output>` — the flat image consumed by an emulator, flash cart, or device.
- `<output>.elf` — the complete ELF from the same link, including DWARF when built with `-g`.

Because both artifacts come from the same link, the ELF addresses match the running flat image exactly. A source-level debugger should load the `.elf` companion.

## Change

Adds a source comment in `lld/ELF/Writer.cpp` at the point where the ELF output path receives its `.elf` suffix. The comment connects that redirect to `writeCustomOutputFormat()`, explains the two artifacts, and records why the companion is the correct debugger input.

This is documentation-only: it changes no linker or code-generation behavior and no output bytes.

**On venue:** if you'd rather keep comment-only churn out of `lld/ELF/Writer.cpp` (it's shared surface for upstream merges), I'm happy to move this to the llvm-mos wiki or llvm-mos-sdk docs instead — the fact seems worth recording wherever it best reaches users.


mysterymath (approved):
If it's not already there, please do add this to the llvm-mos wiki. I do agree that this comment is useful, especially for someone looking for where this happens, and the maintenance overhead should be minimal, since this is already llvm-mos specific code in LLD.

PR #584, opened 2026-07-31
[MOS] Fix non-GPR immediate loads in mos-late-opt

wbniv:
On SPC700, LDImm accepts imaginary-register destinations to represent `mov dp, #imm`. `combineLdImm` tracks only A/X/Y; processing an imaginary destination through that path dereferences a null tracking pointer.

Restrict immediate-load folding to GPR destinations in the existing filtering condition. Other instructions use the existing register-invalidation path. The sibling TA handler keeps its original behavior.

The SPC700 MIR regression covers an imaginary-register immediate load and confirms that it does not interfere with folding an adjacent GPR load to TAX.

Validation: MOS CodeGen suite 79 pass, 1 unsupported (Linux, assertions enabled).


wbniv:
Two updates after a post-hoc review of this PR:

1. **7eedb14 hardens the guard.** The bare `continue` was correct only because no current non-GPR `LDImm` destination aliases `A`/`X`/`Y` — an invariant the register file happens to satisfy rather than one the code enforces. The follow-up commit invalidates any tracked GPR the instruction modifies before continuing, mirroring the invalidation the loop already performs for non-`LDImm` instructions. Behaviour-identical for all current inputs (the checks are dead code until something widens `LDImm`'s destination class further); all four `late-opt*.mir` tests pass unchanged, including the transparency case pinning that a `TAX` rewrite still fires across `$rc2 = LDImm 7`. For the record: the `TA` handler at the top of the same loop has the identical switch-then-store shape, but nothing widens `TA`'s destination class, so it is deliberately left untouched.

2. **The description's live-byte claim was corrected.** The original text said eight simultaneously-live bytes was the smallest count that reliably forces a constant into an imaginary register, and that four stayed in the GPRs. Re-measured (MIR captured with `-mllvm -stop-before=mos-late-opt`, fed to an unfixed `llc -run-pass=mos-late-opt`): **three** live bytes crash at every level above `-O0`, and **two** crash at `-Oz`. The 8-variable repro stands as posted — the bug is easier to hit than the original text claimed, not harder.

wbniv:
Updated (`7eedb14` → `3ce98fe`): new commit hardens the sibling `TA` handler in `combineLdImm`, which has the same switch-then-unconditional-store shape the LDImm fix guards — unreachable today (`TA`'s destination class is XY-only), but this function is where "unreachable today" already failed once, so it now falls through to the generic invalidation instead of trusting the switch. No behavior change for any current input; all four late-opt tests green. Body re-synced with the shipped defensive-invalidation guard.

llvm/lib/Target/MOS/MOSLateOptimization.cpp
```diff
@@ -279,10 +279,17 @@ bool MOSLateOptimization::combineLdImm(MachineBasicBlock &MBB) const {
...
-      // Copy A value to X or Y.
-      Load->MI = &MI;
-      Load->Val = LoadA.Val;
-      continue;
+      // TA's destination class is XY-only today, so Load is always set here.
+      // The guard is defensive: this function is where an "impossible"
+      // destination already appeared once (LDImm below, widened to Anyi8 on
+      // SPC700), so don't trust the switch to be exhaustive. An unhandled
+      // destination falls through to the generic invalidation.
+      if (Load) {
```
mysterymath: Nah, no defensive code of this style please. We already have an effective assert on Load being nullptr via segfault, and crashing is better than silently walking into unexpected territory.

llvm/lib/Target/MOS/MOSLateOptimization.cpp
```diff
@@ -299,6 +306,25 @@ bool MOSLateOptimization::combineLdImm(MachineBasicBlock &MBB) const {
...
+
+    // LDImm's destination is not always a GPR. MOSInstrInfo::getRegClass
+    // widens it to Anyi8 on SPC700, where `$rcN = LDImm imm` is how
+    // `mov dp, #imm` is modelled. Only A, X and Y take part in the rewrites
+    // below; an imaginary destination writes none of them, so there is
+    // nothing to rewrite. Invalidate any tracked GPR the instruction does
+    // modify so a future widening of LDImm's destination class cannot turn
+    // this skip into a stale-value rewrite (no such destination exists
+    // today; these checks are expected to do nothing).
+    if (!MOS::GPRRegClass.contains(Dst)) {
```
mysterymath: Shouldn't this just be a part of the condition above? Might take a teeny bit of refactoring, but that's the right shape, and it makes this huge comment unnecessary.

PR #585, opened 2026-08-04
[65CE02] Legalize arithmetic right shifts

mlund:
Represent one-bit arithmetic right shifts as G_ASHRE during legalization, rather than recovering the sign-carry idiom in instruction selection. On 65CE02, a live result selects native ASR. When store folding makes the result dead, selection retains the existing ROR RMW path using the explicit sign carry. Other CPUs retain the existing CMP #128 + ROR lowering.

Supersedes #552.

#### Byte and cycle impact

| C operation | bytes before | bytes after | Δ | cycles before | cycles after | Δ |
|---|---:|---:|---:|---:|---:|---:|
| `int8_t x = x >> 1` | 4 | 2 | −2 | 10 | 7 | −3 |
| `int8_t x = x >> 2` | 7 | 3 | −4 | 14 | 8 | −6 |
| `int8_t x = x >> 3` | 10 | 4 | −6 | 18 | 9 | −9 |
| `int16_t x = x >> 1` | 12 | 8 | −4 | 25 | 19 | −6 |
| `int16_t x = x >> 2` | 17 | 11 | −6 | 34 | 26 | −8 |
| `int32_t x = x >> 1` | 20 | 12 | −8 | 43 | 29 | −14 |
| `uint8_t x = x >> 1` | 2 | 2 | · | 8 | 8 | · |
| `int16_t g; g >>= 1` (store-folded global) | unchanged | unchanged | · | unchanged | unchanged | · |

Assumes MEGA65 at 40 MHz, 6502 timings for existing instructions, and includes `rts`.

This PR was aided by LLM agent _OpenAI gpt-5.6-terra_ and humanly reviewed in accordance with LLVM guidelines.

**Update**: Using a [patched SDK simulator with 65CE02 opcodes](https://github.com/mlund/llvm-mos-sdk/tree/sim-65ce02), this PR passes the complete [llvm-test-suite](https://github.com/llvm-mos/llvm-test-suite) and all tests in [mega65-freezer](https://github.com/mlund/mega65-freezer) a ~10k line C23 port of the original cc65 code.

### DIFF
--- a/llvm/lib/Target/MOS/MOSCombiner.cpp
+++ b/llvm/lib/Target/MOS/MOSCombiner.cpp
@@ -685,14 +685,20 @@ APInt MOSCombinerImpl::getDemandedBits(Register R,
       DemandedBits |= ~Ones;
       break;
     }
-    case MOS::G_LSHRE: {
+    case MOS::G_LSHRE:
+    case MOS::G_ASHRE: {
       APInt DstDemandedBits = getDemandedBits(MI.getOperand(0).getReg(), Cache);
       if (Use.getOperandNo() == 2) {
         APInt CarryOutDemanded =
             getDemandedBits(MI.getOperand(1).getReg(), Cache);
-        DemandedBits |= DstDemandedBits << 1 | CarryOutDemanded.zext(8);
+        DemandedBits |= DstDemandedBits << 1 |
+                        CarryOutDemanded.zext(DstDemandedBits.getBitWidth());
+        if (MI.getOpcode() == MOS::G_ASHRE)
+          DemandedBits |= DstDemandedBits &
+                          APInt::getSignMask(DstDemandedBits.getBitWidth());
       } else {
         assert(Use.getOperandNo() == 3);
+        // The G_ASHRE LSR/ROR fallback uses carry-in for bit 7.
         DemandedBits |= DstDemandedBits.lshr(7).trunc(1);
       }
       break;
--- a/llvm/lib/Target/MOS/MOSCombiner.cpp
+++ b/llvm/lib/Target/MOS/MOSCombiner.cpp
@@ -538,6 +538,9 @@ bool MOSCombinerImpl::matchFoldShift(MachineInstr &MI,
   if (MI.getOpcode() == MOS::G_LSHRE) {
     CarryOut = (Val->Value & 1).getBoolValue();
     Result = Val->Value.lshr(1) | CarryIn->Value.zext(8).shl(7);
+  } else if (MI.getOpcode() == MOS::G_ASHRE) {
+    CarryOut = (Val->Value & 1).getBoolValue();
+    Result = Val->Value.ashr(1);
   } else {
     CarryOut = (Val->Value & 0x80).getBoolValue();
     Result = Val->Value.shl(1) | CarryIn->Value.zext(8);
@@ -561,7 +564,7 @@ bool MOSCombinerImpl::matchShiftUnusedCarryIn(MachineInstr &MI,
 
   APInt DemandedBits = getDemandedBits(MI.getOperand(0).getReg());
   assert(DemandedBits.getBitWidth() == 8);
-  if (MI.getOpcode() == MOS::G_LSHRE) {
+  if (MI.getOpcode() != MOS::G_SHLE) {
     if ((DemandedBits & 0x80).getBoolValue())
       return false;
   } else {
--- a/llvm/lib/Target/MOS/MOSInstrGISel.td
+++ b/llvm/lib/Target/MOS/MOSInstrGISel.td
@@ -157,6 +157,13 @@ def G_LSHRE : MOSGenericInstruction {
   let InOperandList = (ins type0:$src, type1:$carry_in);
 }
 
+// Arithmetically shift an 8-bit value right one bit. $carry_in must be the
+// source sign bit; it permits a logical-shift fallback when $dst is dead.
+def G_ASHRE : MOSGenericInstruction {
+  let OutOperandList = (outs type0:$dst, type1:$carry_out);
+  let InOperandList = (ins type0:$src, type1:$carry_in);
+}
+
 // These opcodes represent multi-byte increment and decrement operations. Each
 // use operand is one byte; either a register or an absolute address. The def
 // operands correspond to only the register use operands, and they must be tied
--- a/llvm/lib/Target/MOS/MOSCombine.td
+++ b/llvm/lib/Target/MOS/MOSCombine.td
@@ -72,13 +72,13 @@ def fold_sbc : GICombineRule<
 
 def fold_shift : GICombineRule<
   (defs root:$root, build_fn_matchinfo:$matchinfo),
-  (match (wip_match_opcode G_SHLE, G_LSHRE):$root,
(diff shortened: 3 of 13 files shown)


wbniv:
Hi @mlund — I tested #585 against a downstream MOS corpus. The 65CE02 code-size win is real, but I found
one non-65CE02 pessimization caused by an incomplete `G_LSHRE` → `G_ASHRE` migration in demanded-bits
analysis.

`MOSCombinerImpl::getDemandedBits` handles `G_LSHRE` and `G_SHLE`, but not `G_ASHRE`. A `G_ASHRE` use
therefore falls through to `APInt::getAllOnes(Size)`. In a chain of arithmetic shifts, this prevents
`matchShiftUnusedCarryIn` from proving that a preceding shift's bit 7 is dead. On CPUs without native
`ASR`, an otherwise removable `cmp #128; ror` consequently remains instead of becoming `lsr`.

This is safe but pessimistic: demanded bits are over-approximated, so I found no correctness failure.

Minimal C reproducer (`-mcpu=mos6502 -Os -S`):

```c
#include <stdint.h>
typedef struct { int16_t height:5, slope:4, flow:4; uint16_t mat:3; } C;
void step(C *c) { c->flow = (int16_t)(c->flow >> 1); }
```

Before → after #585:

```diff
 	rol
 	cmp	#128
 	ror
-	lsr
-	lsr
+	cmp	#128
+	ror
+	cmp	#128
+	ror
 	lsr
 	and	#30
```

`.text.step` grows 47 → 51 bytes on `mos6502` and 40 → 44 on `mosw65816`. A denser signed-bitfield
kernel grows 132 → 169 bytes on `mos6502` (+28%). A single full-width `a >> 1` does not expose this;
the reproducer needs a shift chain whose high result bits are discarded.

The missing demanded-bits rule is the `G_LSHRE` rule plus source-sign propagation:

```cpp
case MOS::G_ASHRE: {
  APInt DstDemandedBits = getDemandedBits(MI.getOperand(0).getReg(), Cache);
  if (Use.getOperandNo() == 2) {
    APInt CarryOutDemanded =
        getDemandedBits(MI.getOperand(1).getReg(), Cache);
    DemandedBits |= DstDemandedBits << 1 | CarryOutDemanded.zext(8) |
                    (DstDemandedBits & APInt::getSignMask(8));
  } else {
    assert(Use.getOperandNo() == 3);
    DemandedBits |= DstDemandedBits.lshr(7).trunc(1);
  }
  break;
}
```

The sign-mask term is necessary: when destination bit 7 is demanded, `G_ASHRE` also needs source bit 7.
Reusing the plain `G_LSHRE` rule without it would under-approximate demanded bits, which is unsafe.
The carry-in rule remains the same because the non-65CE02 `ROR` fallback consumes carry-in whenever
destination bit 7 is live.

The existing `shift_unused_carry_in` addition does not catch this omission: its `G_ASHRE` result is masked
directly, so `getDemandedBits` never has to traverse through a `G_ASHRE` user. I suggest adding a chained
case such as:

```yaml
%a:_(s8), %ca:_(s1) = G_ASHRE %src, %sign
%b:_(s8), %cb:_(s1) = G_ASHRE %a, %ca
%mask:_(s8) = G_CONSTANT i8 63
%result:_(s8) = G_AND %b, %mask
G_STORE_ABS %result, 1234
```

Without the new arm, the first `G_ASHRE` retains `%sign`; with it, both dead high-bit carry-ins reduce to
zero. That pins the analysis path responsible for the C regression.

I applied the arm locally and rebuilt. The C reducers return byte-for-byte to their pre-#585 output on
`mos6502`, `mos65c02`, and `mosw65816`, while the 65CE02 improvements remain. In a 10,780-object comparison,
all non-65CE02 differences disappeared after the fix; all 16 changed 65CE02 objects retained their wins
(−529 bytes total). A paired emulator differential and the CodeGen/MOS failing set were unchanged.

I also executed the new path rather than only inspecting it. The
[test kernel and xemu harness](https://github.com/wbniv/llvm-mos-65816/tree/d124a9c3190bd5232112062c114a84b9348fba55/dev/c65asr)
are available in my downstream repository. The kernel covers all integer widths and shift amounts, positive
and negative inputs, multi-byte carry chains, store-folded and dead-result forms, and signed-bitfield
read-back. Its host oracle produces checksum `0xE0E8`; bare-metal builds made with the pre-#585 compiler,
#585, and #585 plus the proposed demanded-bits arm all produce the same checksum on xemu's Commodore 65
target. The #585 builds contain 15 native `asr` instructions, compared with none in the baseline, and reduce
the kernel from 1168 to 1068 bytes.

Reproducing that needs no C65 ROM image, which may be worth knowing: xemu's C65 target loads any
128 KiB file as its system ROM without validating it, so the harness just supplies its own.
[xemu](https://github.com/lgblgblgb/xemu) builds in about a minute from source; a prebuilt
`x-emulators` package is also on [apt.foundrylinux.org](https://apt.foundrylinux.org/) if that is
easier.

The reducer, focused MIR regression, and proposed patch are included above.

mlund:
Thanks a lot for looking into this! I find it fairly complex to navigate and predict all side effects. I just pushed the proposed fix:

- Merge `G_LSHRE` and `G_ASHRE` demanded-bits cases with ASR sign-bit dependency
- Used destination width rather than hard-coded 8
- short comment for `G_ASHRE` `LSR`/`ROR` fallback: carry-in supplies bit 7 there.
- Added two missing tests for carry-out to source bit 0 and carry-in demand propagation. Together with the chained sign-bit test, they cover all three terms.
- Added `ASR` to the accumulator-preference/coalescing handling alongside `ASL`/`LSR`/`ROL`/`ROR`

The split-CSR `INW` issue is pre-existing and this branch doesn't incl. `2c516a5`: this branch should rebase after it lands. I left the carry-in invariant documented at the `G_ASHRE` definition, the selector fallback, and now the demanded-bits use.

llvm/lib/Target/MOS/MOSCombiner.cpp
```diff
@@ -561,7 +564,7 @@ bool MOSCombinerImpl::matchShiftUnusedCarryIn(MachineInstr &MI,
 
   APInt DemandedBits = getDemandedBits(MI.getOperand(0).getReg());
   assert(DemandedBits.getBitWidth() == 8);
-  if (MI.getOpcode() == MOS::G_LSHRE) {
+  if (MI.getOpcode() != MOS::G_SHLE) {
```
mysterymath, line 567: Let's reverse the condition here by swapping the if/else bodies: `== G_SHLE` vs else is clearer.

llvm/lib/Target/MOS/MOSInstrGISel.td
```diff
@@ -157,6 +157,13 @@ def G_LSHRE : MOSGenericInstruction {
   let InOperandList = (ins type0:$src, type1:$carry_in);
 }
 
+// Arithmetically shift an 8-bit value right one bit. $carry_in must be the
+// source sign bit; it permits a logical-shift fallback when $dst is dead.
```
mysterymath, line 161: The least surprising shape for `G_ASHR` to be in has a source, destination, and carry out. Any other shape would require significant justfication, and a possible optimization when the destination is dead doesn't seem to be it. After all, if the destination is detected to be dead, can't a combiner just rewrite this into a more efficient code pattern

Issue #66, opened 2021-06-27
Add interrupt annotations


mysterymath:
Clang annotations should be added to support interrupt handling on MOS.

For interrupts to work with the static stack optimization, two function attributes are needed: `__attribute__((interrupt))` and `__attribute__((interrupt_norecurse))`. It will be undefined behavior for an interrupt to call a C function that does not bear one of these two annotations. The only real difference between the two is that if a an interrupt enters a `interrupt_norecurse` function, it is undefined behavior for another interrupt to enter that function while the first invocation is active. `interrupt` has no such requirements. These correspond to roughly the "fast" and "slow" portions of interrupt handlers.

These attributes will cause all necessary ISR save and restore code to be inserted into the annotated function, and the function will return with `RTI` instead of `RTS`. This can be skipped by adding another attribute, `__attribute__((no_isr))`. This indicates that the function is the C entry point for an assembly-language ISR that manages saving and restoring registers. Accordingly, the regular C calling convention will be used, but the other semantics of `interrupt` and `interrupt_norecurse` still apply. It shall be an error to apply this attribute to a function that is not marked as `interrupt` or `interrupt_norecurse`.

Use of any of these attributes anywhere in a program will globally modify the calling convention and use of zero page registers. The number of available zero page pointers will be capped at 5. This ensures that the ISR can execute in a timely fashion, as it must then save and restore 10 - 4 = 6 bytes of zero page pointers (4 bytes are callee-saved and don't need to be saved in ISRs), as well as 5 compiler temporary locations, for a total of 11 LDA PHA pairs, along with the usual pushing of A, X, and Y.

The MOSRecurse pass will need to be modified to walk the call graphs rooted at `interrupt` and `interrupt_norecurse` functions. The following are marked as possibly recursive:

- Any function possibly callable by an `interrupt` function.
- Any function callable by main and an `interrupt_norecurse` function, or by two different `interrupt_norecurse` functions.

This is far from ideal, but it's a starting point. Judicious use of `interrupt_norecurse` will allow carefully designed interrupt handlers to use the static stack. This, along with 5 zero page pointers, should be enough to allow writing working, but highly suboptimal, code.

Later optimization work will improve usage of zero page registers in the presence of C interrupt handlers.

johnwbyrd:
I keep thinking about this problem from the following perspective: what would I do, if I needed to write an interrupt in C, and the 6502 C compiler didn't support interrupt concepts?

[...]

PR #586, opened 2026-08-04
[MOS] Accept an optional BRK signature operand

wbniv:
Accept `brk #imm` as assembler syntax for the BRK opcode followed by a signature byte. Bare `brk` continues to emit one byte.

The signature form is assembler-only. Disassembly always decodes BRK as an operandless, one-byte instruction and treats the following byte independently.

Prior art: ca65 documents an optional BRK signature byte (`brk`, `brk $34`, `brk #$34`) and emits one byte when it is omitted ([ca65 manual, section 4](https://cc65.github.io/doc/ca65.html#ss4.1)). This change accepts only the `#imm` spelling.

Tests cover encoding, object disassembly on MOS6502 and W65816, and disassemble/reassemble round trips. The signature byte in the round-trip test is the NOP opcode, so consuming it as part of BRK would fail the test.

Validation: all 40 MOS MC tests pass (Linux, assertions enabled).

### DIFF
--- a/llvm/lib/Target/MOS/MOSInstrInfo.td
+++ b/llvm/lib/Target/MOS/MOSInstrInfo.td
@@ -192,6 +192,10 @@ def BEQ_Relative : ConditionalBranch<"beq", 0b11, 0b1>;
 /// 8A  9A  AA  BA  CA  EA
 
 def BRK_Implied: InstLow0<"brk", 0b0000>, InstUnconditionalBranch;
+def BRK_Immediate: Inst16<"brk", Opcode<0x00>, Immediate>,
+    InstUnconditionalBranch {
+  let isAsmParserOnly = true;
+}
 /// JSR is the only instruction that does not follow the current pattern.
 def JSR_Absolute: Inst24<"jsr", Opcode<0x20>, Absolute>, InstCall;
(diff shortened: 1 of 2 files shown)


johnwbyrd:
The major concern about such a change is how the disassembler would deal with it.  Strictly, it changes disassembly from being deterministic to non-deterministic.  I could buy off on this, if disassembly always ends up being the Implied version.  Would be nice to see some round-trip tests added to prove that this breaks no one else.

mysterymath:
> The major concern about such a change is how the disassembler would deal with it. Strictly, it changes disassembly from being deterministic to non-deterministic. I could buy off on this, if disassembly always ends up being the Implied version. Would be nice to see some round-trip tests added to prove that this breaks no one else.

I'd be curious what the prior art is here; are other disassemblers (particularly da65) willing to infer a tag for `BRK`? In the absence of compelling prior art, I'd expect `brk #66` to be purely syntactic sugar in the assembler for `brk .byte 66`. I'd never expect the disassembler to produce a `brk #66` accordingly; it's not in the business of lifting the disassembly to a higher level.

Along those lines, I'm broadly in favor, if just because `ca65` supports this.

mysterymath (commented):
This does need a test, as @johnwbyrd mentioned, to establish the disassembler behavior.

PR #588, opened 2026-08-05
[MOS] Add the COP mnemonic with a mandatory signature operand

wbniv:
Add the W65816 COP instruction with a mandatory signature operand, following WDC assembly syntax. COP encodes as opcode $02 followed by the signature byte and disassembles as a two-byte instruction.

Tests cover expression and literal operands, instruction encoding and disassembly, rejection on an unsupported CPU, and rejection of a missing operand. Comments describe the operand requirement without repeating instruction semantics.

Validation: all 40 MOS MC tests pass (Linux, assertions enabled).

### DIFF
--- a/llvm/lib/Target/MOS/MOSInstrInfo.td
+++ b/llvm/lib/Target/MOS/MOSInstrInfo.td
@@ -785,6 +785,14 @@ def TDC_Implied : InstLowB<"tdc", 0b0111>;
 def TSC_Implied : InstLowB<"tsc", 0b0011>;
 
 def WDM_Immediate : Inst16<"wdm", Opcode<0x42>, Immediate>;
+
+/// COP invokes the co-processor software interrupt. Like BRK, it advances the
+/// PC by two, so the byte following the opcode is a "signature" the handler
+/// can read via the pushed return address; the hardware makes no direct use
+/// of it. Unlike BRK, WDC's own assembly syntax requires this operand, so it
+/// is mandatory here rather than optional. W65816-only: on the 65EL02,
+/// opcode $02 is NXT (a different instruction), not COP.
+def COP_Immediate : Inst16<"cop", Opcode<0x02>, Immediate>, InstUnconditionalBranch;
 }
 
 } // Predicates = [HasW65816]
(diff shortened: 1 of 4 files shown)


llvm/lib/Target/MOS/MOSInstrInfo.td
```diff
@@ -785,6 +785,14 @@ def TDC_Implied : InstLowB<"tdc", 0b0111>;
 def TSC_Implied : InstLowB<"tsc", 0b0011>;
 
 def WDM_Immediate : Inst16<"wdm", Opcode<0x42>, Immediate>;
+
+/// COP invokes the co-processor software interrupt. Like BRK, it advances the
```
mysterymath: This is a bit verbose; we haven't taken to describing the semantics of the instructions here; one can look this up elsewhere. Only "Unlike BRK, WDC's own assembly syntax requires this operand, so it is mandatory here rather than optional." actually seems useful.

llvm/test/MC/MOS/asm-errors.s
```diff
@@ -24,3 +24,11 @@ rep #$30
 
 ; CHECK: [[#@LINE+1]]:1: error: instruction requires: FeatureW65816Or65EL02
 sep #$30
+
+; CHECK: [[#@LINE+1]]:1: error: instruction requires: FeatureW65816
+cop #90
+
+; cop has no zero-operand form at all outside FeatureW65816, so a bare
```
mysterymath: Why "outside FeatureW65816"? It doesn't seem like the shape of the instruction depends on the features set.

llvm/test/MC/MOS/cop-signature.s
```diff
@@ -0,0 +1,22 @@
+; RUN: llvm-mc -triple mos -mcpu=mosw65816 -motorola-integers --filetype=obj -o=%t.obj %s
+; RUN: llvm-objdump -d --mcpu=mosw65816 %t.obj | FileCheck %s
+
+; RUN: not llvm-mc -triple mos -mcpu=mos65el02 -motorola-integers %s 2>&1 | FileCheck --check-prefix=EL02 %s
+
+; COP is W65816-only: opcode $02 is NXT on the 65EL02 (a different
```
mysterymath: It shouldn't be expected that W65816 and 65EL02 are the same target; it's odd to keep mentioning that it's not present on the 65EL02. It's also not present on X86 lol.

PR #589, opened 2026-08-05
[MOS] Mark CmpZero as a terminator

wbniv:
Mark CmpZero as a terminator so the machine verifier rejects ordinary instructions placed after it. Restrict `lowerCmpZeros` to the block's terminator range.

Keep each pseudo's fold decision separate from the pass's change accumulator, and report a change when deleting a dead comparison. Tests cover the rejected non-terminator placement, adjacent comparisons that require independent folding/lowering, and removal of a dead comparison. Existing late-optimization tests cover the ordinary single-comparison paths.

The original PR's reproducer placed a normal instruction after CmpZero. That placement violates the intended invariant; this revision encodes and tests the invariant explicitly.

Validation: MOS CodeGen suite 80 pass, 1 unsupported (Linux, assertions enabled).

### DIFF
--- a/llvm/lib/Target/MOS/MOSLateOptimization.cpp
+++ b/llvm/lib/Target/MOS/MOSLateOptimization.cpp
@@ -94,18 +94,25 @@ bool MOSLateOptimization::lowerCmpZeros(MachineBasicBlock &MBB) const {
 
     if (MI.allDefsAreDead()) {
       MI.eraseFromParent();
+      Changed = true;
       continue;
     }
 
     Register Val = MI.getOperand(0).getReg();
 
+    // Per-CmpZero, NOT loop-carried: whether THIS pseudo was folded into an
+    // earlier NZ producer. Sharing `Changed` here made every CmpZero after the
+    // block's first fold skip its lowering too, and nothing downstream lowers
+    // the pseudo -- it reached the asm printer and was emitted as nothing,
+    // silently dropping the flag test.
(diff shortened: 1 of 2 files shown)


mysterymath (changes requested):
Ah, this scenario is explicitly impossible, just for a reason that isn't checked. CmpZero should be a terminator, but it isn't. So it seems the right move here is... to make it a terminator. This also should allow improving the iteration here to just scan terminators.

wbniv:
This is exactly the fix pushed here: `CmpZero` is now `isTerminator = true`, and `lowerCmpZeros` scans `MBB.terminators()` instead of the whole block. The original reproducer (a normal instruction after `CmpZero`) is no longer reachable at all — the verifier itself rejects it, so it's replaced with a negative test (`cmpzero-terminator-invalid.mir`) that asserts exactly that rejection, plus a positive test with adjacent `CmpZero`s that must fold and lower independently.

While making the pseudo a terminator, I also found and fixed the bug the original PR's reproducer was standing in for: `lowerCmpZeros` shared the function's `Changed` accumulator as the per-`CmpZero` "did this one fold" flag, so once any `CmpZero` in a block folded, every later one took the early continue and was never lowered — silently dropping a flag test, since a surviving `CmpZero` is legal MIR and the asm printer emits nothing for it. That's a separate `Folded` flag now, with `Changed` set correctly on the dead-comparison erase path too.

Updated branch incoming.

PR #590, opened 2026-08-05
[MOS] Make zero page allocation deterministic

wbniv:
Zero-page allocation can choose different globals for identical inputs when candidate scores tie. Pointer-ordered iteration also affects entry-point ordering and floating-point accumulation of candidate benefits.

Use insertion-ordered containers at the three relevant iteration sites. Document the determinism requirement without promising a particular iteration order.

The regression gives eight equally beneficial globals four bytes of zero page and checks a consistent allocation. Independent downstream testing in this thread also confirms reproducible full-LTO binaries with the change.

Validation: MOS CodeGen suite 79 pass, 1 unsupported; 20 separate llc runs produce byte-identical assembly.

### DIFF
--- a/llvm/lib/Target/MOS/MOSZeroPageAlloc.cpp
+++ b/llvm/lib/Target/MOS/MOSZeroPageAlloc.cpp
@@ -20,8 +20,10 @@
 #include "MOSMachineFunctionInfo.h"
 #include "MOSRegisterInfo.h"
 #include "MOSSubtarget.h"
+#include "llvm/ADT/MapVector.h"
 #include "llvm/ADT/PostOrderIterator.h"
 #include "llvm/ADT/SCCIterator.h"
+#include "llvm/ADT/SetVector.h"
 #include "llvm/Analysis/BasicAliasAnalysis.h"
 #include "llvm/Analysis/BlockFrequencyInfo.h"
 #include "llvm/Analysis/CallGraph.h"
@@ -414,7 +416,11 @@ SCCGraph MOSZeroPageAlloc::buildSCCGraph(Module &M) {
   LLVM_DEBUG(CG.dump());
 
   std::vector<SCC> SCCs;
-  std::vector<SmallSet<const CallGraphNode *, 4>> SCCCallees;
+  // SmallSetVector, not SmallSet: past its inline capacity SmallSet degrades to
+  // std::set, which orders const CallGraphNode * by pointer VALUE. SCC::Callees
+  // seeds the EntryGraph list, and entry graphs are served ZP round-robin, so an
+  // address-ordered callee list makes the assignment differ between runs.
+  std::vector<SmallSetVector<const CallGraphNode *, 4>> SCCCallees;
   DenseMap<const CallGraphNode *, size_t> SCCIdx;
   std::vector<std::unique_ptr<Candidate>> Candidates;
   DenseMap<GlobalVariable *, Candidate *> GVCandidates;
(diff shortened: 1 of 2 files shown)


andymccall:
I have independently tested both the issue raised and the fix (cc9f0d02), from a different codebase, I've vot three affected targets (`mos-mega65`, `mos-cx16`,`mos-neo6502`), all full-LTO with `-mlto-zp` that I'm compiling a multi-platform gaming API to.

I hit this as a reproducible-build failure on the MEGA65 target - identical source and byte-identical LTO IR (`--save-temps`, all four `.bc` stages hashed) produced 5 distinct binaries across 12 links, sizes spread ~200 bytes on a 37 KB program. `-mllvm -print-after-all` on different runs identified the first differnece to "Allocate zero page (mos-zero-page-alloc)" independently reaching the same pass this issue names. I then found the same three sites in `MOSZeroPageAlloc.cpp` before finding this issue, treat this as similar diagnosis rather than confirmation bias, as I hadn't seen this issue until after I'd found the cause.

Verification of cc9f0d02 at the pinned parent of my toolchain (c798c31416f7), red first, then green:

- your 8-tied-globals reproducer, unpatched llc: 6 distinct winner sets in 20 runs. Patched: 20/20 identical, winners in module order as predicted.
- 37 KB whole-program LTO link, unpatched, ASLR on: 5 distinct binaries in 12 links. Patched: 16/16 identical, and 8/8 through the full compiler driver with its randomly named /tmp objects.
- a second consumer's `mos-cx16` target went from characterised-bimodal (two independent projects measured 10-byte and 69-byte mode gaps) to 8/8 identical.
- the patched output was byte-identical to one of the tickets the unpatched linker had already produced, on two different trees - ie. the fix picks a canonical member of the existing output family rather than changing
  codegen.

One additional data point supporting the temp-name observation in the issue: with ASLR disabled and byte-identical inputs, renaming a single file that's an input object (same path length, two characters changed mid-name) flipped the unpatched output. With cc9f0d02 applied, the renamed arms remained byte-identical - so the name sensitivity routes through the same pointer-order tie-breaks and dies with them.

Happy to share the full A/B logs if useful. This fix is carrying my production builds now; it would be good to see it in a release so.

llvm/lib/Target/MOS/MOSZeroPageAlloc.cpp
```diff
@@ -414,7 +416,11 @@ SCCGraph MOSZeroPageAlloc::buildSCCGraph(Module &M) {
   LLVM_DEBUG(CG.dump());
 
   std::vector<SCC> SCCs;
-  std::vector<SmallSet<const CallGraphNode *, 4>> SCCCallees;
+  // SmallSetVector, not SmallSet: past its inline capacity SmallSet degrades to
+  // std::set, which orders const CallGraphNode * by pointer VALUE. SCC::Callees
+  // seeds the EntryGraph list, and entry graphs are served ZP round-robin, so an
+  // address-ordered callee list makes the assignment differ between runs.
+  std::vector<SmallSetVector<const CallGraphNode *, 4>> SCCCallees;
```
mysterymath, line 422: Just "Note: iteration order must be deterministic.". Here and throughout.

llvm/test/CodeGen/MOS/zp-alloc-deterministic.ll
```diff
@@ -0,0 +1,59 @@
+; RUN: llc -mtriple=mos -zp-avail=4 -verify-machineinstrs < %s | FileCheck %s
+
+; Zero page allocation must be deterministic.
+;
+; Every one of these globals is stored exactly once from the same basic block,
+; so MOSZeroPageAlloc computes the identical benefit (2 * 1.0 / 1) for all eight
+; and only four of them fit in the zero page. The winners are therefore decided
+; entirely by the order in which candidates were collected, which must be the
```
mysterymath: which must be deterministic. Let's not specify the exact order: it doesn't matter.

llvm/test/CodeGen/MOS/zp-alloc-deterministic.ll
```diff
@@ -0,0 +1,59 @@
...
+; MachineInstr walk order -- i.e. module order, g0 through g3.
+;
+; Before the fix, collectCandidates() accumulated benefits in a
```
mysterymath: Noone cares how code was broken in the past; remove this whole paragraph.

wbniv:
Thanks — this is exactly the kind of independent confirmation the PR needed, and the renamed-input experiment is a nice sharpening of the diagnosis: with ASLR off, a file name is one of the few things that still perturbs heap layout, so it reaching the output through the same pointer-order tie-breaks and vanishing with the fix is what the analysis predicts.

Yes please, share the A/B logs. A gist link or an attachment on this PR is fine; if it's convenient, the unpatched 12-link spread (sizes and first-differing pass) and the 8/8 post-fix driver runs are the two I'd most like to have on record here. I'll cite them from the PR description.

A revision addressing the review comments above (comment wording only, no code change) is queued and will land on this branch shortly.

mysterymath (approved):
LGTM!

PR #549, opened 2026-02-09
[65CE02] Fix 16-bit branch PC-relative offset encoding

mlund:
## Summary
- Fix `getRelativeMOSPCCorrection()` to use -1 correction for 65CE02 16-bit branches (PC+2 base), keeping -2 for 65816 BRL (PC+3 base)
- The 65CE02 computes branch target as `opcode_addr + 2 + offset` for both 8-bit and 16-bit forms; the assembler incorrectly used `opcode_addr + 3` (the 65816 convention)
- Query `Feature65CE02` from the fragment's `MCSubtargetInfo` to distinguish architectures
- Add backward branch, boundary (127/128), and 45GS02 coverage to MC test

## References
Confirmed against three independent sources:
- MEGA65 FPGA core (`gs4510.vhdl`, `B16TakeBranch` microcode)
- Xemu emulator (`cpu65.c`, `_BRA16` function)
- [Pagetable.com 65CE02 reference](https://pagetable.com/c64ref/6502/cpu_65ce02.txt)

## Test plan
- [x] `llvm-lit llvm/test/MC/MOS/long-branches-65ce02.s` — updated encoding + edge cases
- [x] `llvm-lit llvm/test/MC/MOS/long-branches-65816.s` — no regression for 65816 BRL
- [x] Full MOS MC + CodeGen test suite (117/117 pass)
- [x] vector_logo.prg runs correctly in xmega65 emulator
- [x] **Update**: Using a [patched SDK simulator with 65CE02 opcodes](https://github.com/mlund/llvm-mos-sdk/tree/sim-65ce02), this PR passes the complete [llvm-test-suite](https://github.com/llvm-mos/llvm-test-suite) and all tests in [mega65-freezer](https://github.com/mlund/mega65-freezer) a ~10k line C23 port of the original cc65 code.

Co-authored with Claude AI and humanly supervised and reviewed by @mlund in accordance with LLVM guidelines.

PR #591, opened 2026-08-05
[MOS] Report an error on out-of-range PCRel8/PCRel16 branch fixups

wbniv:
## Summary

`MOSAsmBackend::applyFixup` applies the PC-relative correction to `PCRel8`/`PCRel16` branch
fixups and writes the result without checking that it fits: a branch whose target is out of
range assembles silently, with the offset truncated into the encoding — the branch lands
somewhere else, with no diagnostic at any stage.

This patch checks resolved `PCRel8`/`PCRel16` values against their ranges after the
correction is applied, and reports a source-located error when they don't fit:

```
error: 8-bit branch target out of range
error: 16-bit branch target out of range
```

`llvm/test/MC/MOS/branch-range-errors.s` pins both messages; it is red without the change (the
offset truncates silently, so no diagnostic is emitted) and green with it. The suites are
otherwise unchanged: `llvm/test/MC/MOS/` 40/40, `llvm/test/CodeGen/MOS/` 78 pass + 1
upstream-disabled, 0 failures.

## Relation to #549

#549 fixes the *value* of the PC-relative correction per CPU (65CE02 16-bit branches are
PC+2-based, unlike the 65816's PC+3 `BRL`); this patch adds *range checking* of the corrected
value. They touch the same few lines and are designed to compose: whichever lands first, the
other rebases trivially, and the diagnostics then guard #549's corrected offsets too.


wbniv:
Added a regression net for the #549 class of bug this PR's range check would otherwise miss: boundary-exact legal/illegal pairs at the PCRel8/PCRel16 edges, and a byte-exact resolved-fixup matrix across mos6502/mosw65816/mos65ce02, all hand-derived from the datasheet rather than captured from the assembler. The 65CE02 16-bit rows are written with the datasheet-correct values `#549` restores and are `XFAIL`'d against today's still-buggy correction constant — they'll flip to passing (and lose the `XFAIL`) once this rebases onto #549, which is a nice joint proof that the two PRs compose correctly.

llvm/test/MC/MOS/branch-range-matrix-65ce02-16bit-pending-549.s
johnwbyrd, line 1: All these comments scream AI-generated without AI attribution.  I'd like to see a clean version of this without the running commentary.

johnwbyrd (commented):
https://llvm.org/docs/AIToolPolicy.html

wbniv:
Thanks—I removed the running commentary and retained only concise boundary labels and the 65CE02 PC-relative invariant needed to understand the tests. I also rebased onto current main; since #549 has merged, the corresponding 65CE02 test is now enabled.

Validation: the updated llvm-mc and llvm-objdump built successfully, and all five focused MC test files passed their exact RUN pipelines.

Disclosure: I used an AI coding assistant while developing and revising this change. I reviewed and validated the resulting code and tests.

mysterymath (approved):
LGTM.

PR #551, opened 2026-02-18
[65CE02] Chained INW/DEW for small constant offsets

mlund:
## Summary

- Emit chained `G_INC/G_DEC` for small constant pointer and integer additions on 65CE02+
- Post-RA, each `G_INC/G_DEC` becomes INW/DEW when register allocator places bytes in adjacent ZP pairs
- Asymmetric thresholds: ptr ±4, integer increment +2, integer decrement -1

Motivating pattern: strength-reduced `ptr[i*2]` loop access (e.g. `fire_fcm.cc`) becomes `ptr += 2` per iteration → `inw zp; inw zp` (4 bytes) instead of `lda/clc/adc #2/sta/lda/adc #0/sta` (14 bytes).

## Before/after (45GS02 vs 6502, typical cycles, INW/DEW = 7 cyc)

| Operation | 6502 bytes | 45GS02 bytes | 6502 cyc | 45GS02 cyc |
|-----------|-----------|-------------|---------|-----------|
| i16 += 2  | 10        | 9  (-1)     | 20      | 16  (-4)  |
| ptr += 2  | 14        | 5  (-9)     | 24      | 20  (-4)  |
| ptr += 3  | 14        | 7  (-7)     | 24      | 27  (+3)  |
| ptr += 4  | 14        | 9  (-5)     | 24      | 34 (+10)  |
| ptr -= 2  | 14        | 5  (-9)     | 24      | 20  (-4)  |
| ptr -= 4  | 14        | 9  (-5)     | 24      | 34 (+10)  |
| i32 += 2  | 22        | 17  (-5)    | 36      | 16 (-20)  |

## Real-world: `fire_fcm.cc` flame effect in full color mode on MEGA65

Chained `inw zp; inw zp` confirmed at two hot sites (stride-2 pointer access):
- `setup_screen`: screen pointer += 2 (1000 iterations)
- `convert_and_dma`: fire row pointer += 2 (1000 iterations/frame)

Per-site savings: 10 bytes, 5 instructions, 10 cycles per iteration.
Per frame: ~20,000 fewer cycles across both sites

Screenshot 2026-02-18 at 17 54 56

## Test plan

- [x] 77/77 MOS CodeGen tests pass
- [x] No change to 6502 codegen (thresholds gated on `has65CE02()`)
- [x] Edge cases: sub, add 0, i64, i8, boundary values
- [x] Verified chained INW in real-world `fire_fcm.cc` hot loops
[...]

PR #593, opened 2026-08-05
[65CE02] Keep late `INW`/`DEW` out of split ZP-stack register pairs

### LINKED ISSUE #592: [65CE02] `DEW` can update the wrong loop counter with LTO
With LTO and `-mlto-zp` > 4, a 16-bit loop counter can be moved to the zero-page stack. The loop test then reads that stack slot, but `DEW` still updates the original imaginary register. The loop counter never changes from the test's point of view. A bounded loop can therefore run past the end of an array or never finish. Found during compilation of the mega65 freezermenu (>10000 lines of C) which applies significant pressure.

## Reproducing

Minimal example attached as `repro.c`.

```sh
mos-mega65-clang -Oz -flto -mcpu=mos45gs02 -mlto-zp=210 \
  -Wl,--lto-emit-asm -o out.prg repro.c
```

generated assembly is in `out.prg.lto.s`.

## Wrong output

The loop test reads the stack slot:

```asm
lda mos8(.Lmain_zp_stk+1)
...
lda mos8(.Lmain_zp_stk)
```

But the decrement updates a different location:

```asm
dew __rc20
```

`DEW` should instead use the same stack slot as the loop test, for example:

```asm
dew mos8(.Lmain_zp_stk)
```

## Affected version

Observed with:

```text
clang version 23.0.0git (llvm-mos c798c31416f72b395c658b5502d281a162387ab1)
```

reproduces with `-mlto-zp=8` through `210`; `-mlto-zp=0` and `4` work.

## Not-so-nice workaround

Use `-mlto-zp=4` or lower, or compile for `mos6502`, or use clang prior to #547 #548

mlund:
Fixes a miscompile on 65CE02 targets when LTO moves the two bytes of a counter to the zero-page stack. `FinalizeISel` can later replace byte operations with `INW` or `DEW`. If the pair has no matching 16-bit stack entry, the instruction used the original imaginary register while the rest of the code used the stack slot. Do not form `INW` or `DEW` for such a split pair. The existing byte operations use the right stack locations.

Also add an assertion that catches a missing 16-bit stack entry during assembly printing.

Tests:

- New 65CE02 and 45GS02 LTO-style test with a counter in the zero-page stack.
- Existing MOS code-generation tests.
- Corrected flawed codegen in the mega65 freezemenu (>10000 lines of C) as reported in original issue.
- **Update**: Using a [patched SDK simulator with 65CE02 opcodes](https://github.com/mlund/llvm-mos-sdk/tree/sim-65ce02), this PR passes the complete [llvm-test-suite](https://github.com/llvm-mos/llvm-test-suite) and all tests in [mega65-freezer](https://github.com/mlund/mega65-freezer) a ~10k line C23 port of the original cc65 code.

Fixes #592 and should land after #551.

The _GPT 5.6 Terra_ coding agent assisted in creating this patch, which is humanly reviewed and tested by @mlund in accordance with LLVM guidelines.


llvm/lib/Target/MOS/MOSLegalizerInfo.cpp
```diff
@@ -672,22 +672,26 @@ bool MOSLegalizerInfo::legalizeAddSub(LegalizerHelper &Helper,
...
 
-  // Handle multi-byte increments and decrements.
-
-  assert(MI.getOpcode() == MOS::G_ADD || MI.getOpcode() == MOS::G_SUB);
-  int64_t Amt = RHSConst->Value.getSExtValue();
-  if (MI.getOpcode() == MOS::G_SUB)
-    Amt = -Amt;
+  // Take the magnitude without negating; -INT64_MIN would overflow.
+  const int64_t Amt = RHSConst->Value.getSExtValue();
+  if (llvm::AbsoluteValue(Amt) != 1)
```
mysterymath, line 686: AFAICT this doesn't fix anything; `llvm::AbsoluteValue` cannot return anything sensible for INT64_MIN either. That means that we need a 3-way compare with +1 and -1 here. And it's hard to find a meaningful comment to explain *why* without referencing a shape the code doesn't have anymore.

Accordingly, I think we can skip breaking this out into a SExtValue entirely and just express this on the APInt. Easier to hold.

llvm/lib/Target/MOS/MOSLegalizerInfo.cpp
```diff
@@ -1471,9 +1475,26 @@ bool MOSLegalizerInfo::legalizePtrAdd(LegalizerHelper &Helper,
...
 
-  if (ConstOffset && ConstOffset->Value.abs().isOne()) {
-    Builder.buildInstr(ConstOffset->Value.isOne() ? MOS::G_INC : MOS::G_DEC,
-                       {Result}, {Base});
+  // Pointers are allocated to Imag16, so the byte pair is always adjacent and
+  // each step becomes a single INW/DEW. The limit is cycles, not allocation:
+  // one INW/DEW beats a 16-bit ADC sequence, two roughly break even on speed
+  // while saving space, and longer chains lose. Measured on a MEGA65, raising
+  // this to 4 cost 3% in a tile-conversion loop.
+  const MOSSubtarget &STI = Builder.getMF().getSubtarget<MOSSubtarget>();
```
mysterymath, line 1483: This seems to change codegen beyond what is mentioned in the description. Am I wrong about that?

llvm/lib/Target/MOS/MOSMCInstLower.cpp
```diff
@@ -862,6 +862,11 @@ bool MOSMCInstLower::lowerOperand(const MachineOperand &MO, MCOperand &MCOp) {
       break;
     }
 
+    assert((!MOS::Imag16RegClass.contains(Reg) ||
```
mysterymath, line 865: Why is it a problem if the entire pair is relocated to the stack? Do we not guarantee they are contiguous? It seems like we probably could?

llvm/lib/Target/MOS/MOSISelLowering.cpp
```diff
@@ -468,10 +469,15 @@ static MachineBasicBlock *emitIncDecMB(MachineInstr &MI,
       WordReg = TRI.getMatchingSuperReg(UseReg, MOS::sublo,
                                         &MOS::Imag16RegClass);
       if (WordReg && TRI.getSubReg(WordReg, MOS::subhi) == NextUseReg) {
+        const auto &CSRZPOffsets =
+            MBB->getParent()->getInfo<MOSFunctionInfo>()->CSRZPOffsets;
+        bool SplitCSR = !CSRZPOffsets.count(WordReg) &&
```
mysterymath, line 474: This expression is hard for me to parse the semantics of. It could use refactoring or a comment.

llvm/lib/Target/MOS/MOSZeroPageAlloc.cpp
```diff
@@ -556,16 +556,19 @@ void MOSZeroPageAlloc::collectCandidates(
...
+    return SavedRegs.test(Reg) &&
+           !MF.getRegInfo().getUsedPhysRegsMask().test(Reg);
+  };
   unsigned Idx = 0;
   DenseSet<Register> Imag16Regs;
   for (Register Reg : SavedRegs.set_bits()) {
     if (!MOS::Imag8RegClass.contains(Reg))
       continue;
 
-    // If a call occurs with a different calling convention, it may clobber
```
mysterymath, line 565: Deleting this comment feels very wrong. Is this not a concern anymore? Why?

llvm/lib/Target/MOS/MOSZeroPageAlloc.cpp
```diff
@@ -578,13 +581,16 @@ void MOSZeroPageAlloc::collectCandidates(
       Benefit += 12 * RestoreFreq; // LDA ABS,STA ZP
     }
 
-    // If the CSR is used as a 16-bit pointer, then the two halves cannot be
```
mysterymath, line 581: Again, it feels wrong to drop this comment.

llvm/lib/Target/MOS/MOSLegalizerInfo.cpp
mlund: Yes, it's from #551 which is now merged. I need to rebase.

PR #595, opened 2026-08-16
[GlobalISel] Zero-extend G_MEMSET's fill value to int

xobs:
G_MEMSET's fill operand is i8, but the C memset() it is lowered to takes int. On targets where the calling convention distinguishes the two (e.g. MOS, where an i8 travels in A but an int occupies two registers), passing the i8 straight through assigns it to the wrong location and memset reads a garbage fill value.

Zero-extend the fill value to the int width (pointer width, capped at 32 bits) so the libcall argument matches the C prototype.

This was a subtle one where `memset()` would silently write the wrong values. Probably should be upstreamed, but really only an issue for targets with very small int sizes like 6502.


xobs:
This had a different root cause

PR #597, opened 2026-08-22
[compiler-rt] MOS hand-tuned single-precision softfloat (2nd attempt)

johnwbyrd:
Round two of the softfloat PR (previous attempt closed at #575).
Rewritten from naked-C to `.S`, with two subtract-path bugs found by
@wbniv fixed, and in-tree regression coverage added for the bug classes
that the existing 25-vector `addsf3_test.c` did not probe.

## Summary

Adds MOS 6502 hand-tuned `.S` implementations of the four single-precision
IEEE 754 soft-float compiler-rt builtins: `__addsf3`, `__subsf3`,
`__mulsf3`, `__divsf3`.  All four confine state to caller-saved zero-page
slots (`__rc2..__rc19`), so there is no prologue/epilogue and no stack
frame.  Rounding is hardcoded round-to-nearest-ties-to-even (no fenv --
the generic compiler-rt fenv machinery is ~30x more expensive per call
than the arithmetic it wraps).

## Numbers (vs upstream generic, on the existing compiler-rt Unit tests, under mos-sim)

| routine   | vectors        | isolated .text gen -> mos | full-test cycles gen -> mos     |
|-----------|----------------|---------------------------|----------------------------------|
| `__addsf3`  | 43             | 3130 -> **703 B** (-78%)  | ~34,000 -> ~35,700               |
| `__subsf3`  | 40             | -> **11 B** (trampoline)  | (delegates to `__addsf3`)          |
| `__mulsf3`  | 350+           | 2424 -> **681 B** (-72%)  | 2,310,376 -> **967,608** (-58%)  |
| `__divsf3`  | 351            | 3305 -> **806 B** (-76%)  | 5,438,847 -> **535,174** (-90%)  |

The 10x `__divsf3` speedup is because the generic path is Newton-
Raphson (three iterations, each needing a 32x32 multiply -- catastrophic
on the 6502).  Shift-subtract long division is a much better algorithmic
fit here.

End-to-end validation via a 60x25 ASCII Mandelbrot demo: **bit-identical
output** vs the generic softfloat implementation (MD5 verified) at
**3.2x speed** (~200M cycles vs ~638M).

## Two subtract-path bugs fixed relative to the earlier naked-C version

Both were discovered by @wbniv transcribing the previous implementation
into a Python model and fuzzing against numpy float32 as ground truth.

- **Bug 1 (severe) -- subnormal-result doubling.**  When the correctly-
  rounded result of `a - b` is a nonzero subnormal float (< ~1.175e-38),
  an implementation whose normalize bit-loop tests `A_EXP == 0` before
  deciding whether to shift does one extra shift-and-decrement at the
  `exp==1 -> exp==0` transition.  IEEE 754 subnormals reuse the same
  reference scale as the smallest normal (they just drop the implicit
  leading 1), so the loop must stop at that boundary without shifting.
  Buggy implementations return exactly 2x the correct magnitude -- in a
  random fuzz, ~31% of nonzero-subnormal difference cases came back
  doubled.
- **Bug 2 -- alignment sticky-bit sign flip.**  On the subtract path,
  when `b` needs an alignment right-shift and some of `b`'s low bits are
  truncated, the computed difference is slightly too *large* by
  <1 ULP.  Correct rounding must bias DOWN -- the mirror of the
  add-path convention where a nonzero sticky biases up.  Implementations
  that reuse the add-path rounding logic on the subtract path produce
  results that are exactly +1 ULP high (~63/300k in random fuzz, always
  d=+1).

Regression vectors covering both classes have been added to
`compiler-rt/test/builtins/Unit/addsf3_test.c` (+18) and a new
`subsf3_test.c` (+40, includes 22 baseline IEEE 754 shape cases).
The vectors are hex-encoded so they exercise the same shapes on any
target implementing the softfloat builtins.

## Response to review feedback from #575

- **@wbniv** identified both subtract-path bugs and suggested the fixes
  used here.  Thank you.
- **@mysterymath** pointed out that naked-C + `asm volatile` was the
  wrong shape.  Confirmed empirically: LTO cannot optimize inside a
  naked+`asm volatile` body, and the MOS toolchain does DCE via
  `--gc-sections` + archive semantics regardless of LTO, so nothing was
  gained by staying in C.  This PR is `.S`.

## Assembler notes

- Local labels use a per-function `.L__addsf3_*` / `.L__mulsf3_*` /
  `.L__divsf3_*` prefix.  Keeps them out of the object symbol table and
  unique across sibling files.
- Far branches (>127 B) use invert-and-jmp or a shared JMP trampoline
  to work around the LLVM MC assembler's silent 8-bit branch-offset
  truncation.
- `.zeropage __rcN` directives at the top of each `.S` declare the
  imaginary registers as zero-page symbols, so `lda`/`sta` emit the
  2-byte zp encoding.  See https://llvm-mos.org/wiki/Assembler_relaxation.

## Test plan

- [x] `Unit/addsf3_test.c` (43 vectors: 25 upstream + 18 regression) passes
- [x] `Unit/subsf3_test.c` (40 vectors: 22 baseline + 18 regression) passes
- [x] `Unit/mulsf3_test.c` (350+ vectors) passes
- [x] `Unit/divsf3_test.c` (351 vectors) passes
- [x] 27,656-vector fuzz against `__subsf3` (spanning cancel/align/full-STK/random shapes): 0 mismatches
- [x] Full compiler-rt build (`ninja runtimes/install-builtins-mos-unknown-unknown`)
- [x] Downstream `llvm-mos-sdk` `libcrt.a` rebuilds cleanly
- [x] End-to-end Mandelbrot via installed archive: bit-identical to generic, 3.2x speed

### DIFF
--- a/compiler-rt/lib/builtins/CMakeLists.txt
+++ b/compiler-rt/lib/builtins/CMakeLists.txt
@@ -938,7 +938,13 @@ set(mips64_SOURCES ${GENERIC_TF_SOURCES}
 set(mips64el_SOURCES ${GENERIC_TF_SOURCES}
                      ${mips_SOURCES})
 
-set(mos_SOURCES ${GENERIC_SOURCES})
+set(mos_SOURCES
+  mos/addsf3.S
+  mos/divsf3.S
+  mos/mulsf3.S
+  mos/subsf3.S
+  ${GENERIC_SOURCES}
+)
 # MOS SDK implementations are already smaller and faster.
 list(REMOVE_ITEM mos_SOURCES
   ashldi3.c
--- a/compiler-rt/lib/builtins/mos/divsf3.S
+++ b/compiler-rt/lib/builtins/mos/divsf3.S
@@ -1,0 +1,858 @@
+//===-- lib/mos/divsf3.S - Single-precision division (MOS 6502) -*- Asm -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// ===----------------------------------------------------------------------===//
+//
+// __divsf3(a, b) -- IEEE 754 binary32 division for MOS (6502).
+//
+// Rounding: round-to-nearest-ties-to-even, hardcoded.  No fenv support.
+//
+// Same overall shape as addsf3.c and mulsf3.c: one naked top-level
+// function with all state in caller-saved zero-page (__rc2..__rc19),
+// .L-prefixed local labels, invert-and-jmp for far branches.  See
+// addsf3.c for the reference material on IEEE 754 layout and the
+// assembler gotchas that applies unchanged here.
+//
+// ============================================================================
+// REFERENCE: division algorithm outline
+// ============================================================================
+//
+// value(a/b) = (-1)^(sa^sb) * (Ma/Mb) * 2^((ea-eb))     for normal a,b
+//
+// Sign of result is the XOR of input signs.  Magnitude is a 24 x 24 ->
+// 25-bit mantissa division plus an exponent subtract.  Since both input
+// mantissas are in [2^23, 2^24) (implicit-1 restored), the mantissa
+// ratio Ma/Mb is in (0.5, 2.0).  Two cases fall out:
+//
+// Case A (Ma >= Mb, ratio in [1, 2)):  the true (Ma * 2^24) / Mb
+// quotient is a 25-bit value with bit 24 = 1.  The result mantissa
+// (with implicit-1) is the top 24 bits of that quotient.
+// Case B (Ma < Mb, ratio in [0.5, 1)):  quotient is a 24-bit value
+// in [2^23, 2^24).  The result mantissa is those 24 bits directly,
+// and the result exponent is decremented by 1 (to compensate for
+// the [0.5, 1) → [1, 2) normalization).
+//
+// Which case applies is determined by a pre-loop 24-bit compare of
+// Ma and Mb; the flag is stored in Q_HI (1 for case A, 0 for case B).
+//
+// The divide loop is a standard 6502 shift-subtract long division.
+// Init: R = Ma; Q = 0.  For case A, we also pre-subtract D once from
+// R (matching the "would-be first quotient bit = 1" of the 25-bit
+// quotient) to establish the loop invariant "R < D".  Case B needs no
+// pre-subtract because Ma < Mb already satisfies it.
+//
+// Then 24 iterations of:
+//
+// 1. Shift [R:Q] left by 1.  The bit that falls off R2's MSB
+// (captured in the C flag as C_out) reflects whether pre-shift R
+// had bit 23 set -- i.e. whether the "true" 25-bit post-shift R
+// was >= 2^24 > D.
+// 2. If C_out = 1:  force subtract D from the 24-bit R.  The 24-bit
+// SBC's borrow "consumes" the 2^24 bit that shifted out, so the
+// stored 24-bit R correctly reflects the 25-bit R minus D.
+// 3. Else:  compare R with D (high byte first, early-out) and
+// subtract in place only if R >= D.
+// 4. If we subtracted (in either branch), set Q bit 0 = 1.
+//
+// A post-loop cleanup handles the exact-division boundary: if R >= D
+// after 24 iterations, subtract D once more and increment Q.  (Without
+// this, exact-divisible cases like Ma == Mb would end with R = D and
+// Q one short.)
+//
+// After the loop, Q holds the 24-bit low bits of the true quotient
+// (which is 24-bit in case B or the low 24 bits of a 25-bit value in
+// case A):
+//
+// Case A: mantissa = 0x800000 | (Q >> 1); round bit = Q bit 0.
+// Case B: mantissa = Q; round bit computed from a "virtual 25th
+// iteration" (shift R left, compare/sub with D, round bit = 1 iff
+// the virtual subtract commits).
+//
+// Sticky = (R != 0) after all shifts and the virtual iteration.
+//
+// Underflow (R_EXP <= 0) triggers a shift-right of the mantissa with
+// careful round/sticky tracking:  the old round bit becomes part of
+// the new sticky before each shift, and the bit that falls off each
+// shift becomes the new round bit.  After (1 - R_EXP) shifts, the
+// mantissa is in subnormal-encoding form and R_EXP is set to 0.
+//
+// ============================================================================
+// REFERENCE: result exponent
+// ============================================================================
+//
+// R_EXP = a_exp + (127 - b_exp)
+// = (a_exp - b_exp) + 127
+//
+// For normal inputs (a_exp, b_exp in [1, 254]) this is in
+// [1-254+127, 254-1+127] = [-126, 380].  Fits in signed 16-bit.
+// Subnormals renormalize like in mulsf3 -- shift the significand left
+// until bit 23 is set, adjust the effective exponent one per shift.
+//
+// After the divide loop's mantissa case-B branch (ratio < 1), R_EXP
+// gets one more decrement.
+//
+// ============================================================================
+// REFERENCE: special cases
+// ============================================================================
+//
+// NaN / anything = quiet(a)
+// anything / NaN = quiet(b)
+// Inf / Inf      = qNaN
+// Inf / finite   = signed Inf
+// finite / Inf   = signed 0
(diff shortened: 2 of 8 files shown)


compiler-rt/lib/builtins/mos/divsf3.S
```diff
@@ -0,0 +1,880 @@
...
+// is consumed as quotient bits shift up from below).
+//
+// ===----------------------------------------------------------------------===//
+
+
+// __rcN are provided by imag-regs.ld at link time in zero page.  Tell the
+// assembler so it emits the 2-byte zp encoding for lda/sta/... instead of
+// the 3-byte absolute form (which the linker cannot relax back down to zp
+// once emitted).  See https://llvm-mos.org/wiki/Assembler_relaxation.
+.zeropage __rc2
```
mysterymath: Seems like we should have an include for this?

compiler-rt/test/builtins/Unit/addsf3_test.c
```diff
@@ -92,5 +92,41 @@ int main(void) {
   // Large normal values near overflow
   status |= test__addsf3(0x7f7fffff, 0x7f7fffff, 0x7f800000); // max + max = inf
 
+  // Regression: opposite-sign near-cancellation whose true result is a nonzero
```
mysterymath, line 95: Is it actually possible to run these? How?

compiler-rt/lib/builtins/mos/divsf3.S
johnwbyrd: Good call. Done in 53f025b — pulled the `.zeropage __rc2..__rc19` block out into [`compiler-rt/lib/builtins/mos/imag-regs-zp.inc`](https://github.com/llvm-mos/llvm-mos/blob/pr/compiler-rt-mos-float/compiler-rt/lib/builtins/mos/imag-regs-zp.inc) and each of the four `.S` files now just `#include "imag-regs-zp.inc"`. Uses the C preprocessor include (which finds sibling headers automatically) rather than assembler `.include` (which would need an extra `-I` on the ASM command line).

compiler-rt/test/builtins/Unit/addsf3_test.c
johnwbyrd: Compiler-rt's lit config uses `%run %t`, which for hosted targets executes the linked binary directly. The mos target has no host binary — the link output is a `mos-sim` ROM image, not a native ELF — so `%run` for mos isn't wired up. **Right now these regression cases are not exercised by mos CI.** They run automatically on every other target that CI exercises (x86, ARM, RISC-V, ...), which is where they'd catch a shape regression in a portable softfloat rewrite. Wiring `mos-sim` into lit's `%run` for the mos target would be a nice follow-up but is a separate infrastructure change.

For now, running them as a one-off on the mos target looks like this (from the llvm-mos source root, with the SDK's install/bin and llvm-mos's install/bin on `$PATH` so `mos-sim-clang` and `mos-sim` resolve):

```sh
CRT=compiler-rt

# addsf3_test — links only the addsf3.S source, plus the test
mos-sim-clang -Os \
    -I "$CRT/lib/builtins" -I "$CRT/test/builtins/Unit" \
    $CRT/lib/builtins/mos/addsf3.S \
    $CRT/test/builtins/Unit/addsf3_test.c \
    -o /tmp/addsf3_test
mos-sim /tmp/addsf3_test ; echo "exit=$?"

# subsf3_test — subsf3 is a trampoline into addsf3, so link both
mos-sim-clang -Os \
    -I "$CRT/lib/builtins" -I "$CRT/test/builtins/Unit" \
    $CRT/lib/builtins/mos/addsf3.S $CRT/lib/builtins/mos/subsf3.S \
    $CRT/test/builtins/Unit/subsf3_test.c \
    -o /tmp/subsf3_test
mos-sim /tmp/subsf3_test ; echo "exit=$?"

# mulsf3_test / divsf3_test — same shape as addsf3_test with the corresponding .S
```

`exit=0` on `mos-sim` means every case passed; any nonzero exit means at least one printed an `error at line N: ...` line to stdout before quitting. That's the loop I've been using while writing this.

compiler-rt/lib/builtins/CMakeLists.txt
```diff
@@ -938,7 +938,20 @@ set(mips64_SOURCES ${GENERIC_TF_SOURCES}
 set(mips64el_SOURCES ${GENERIC_TF_SOURCES}
                      ${mips_SOURCES})
 
-set(mos_SOURCES ${GENERIC_SOURCES})
+set(mos_SOURCES
```
mysterymath, line 941: It looks like there is already an existing mechanism for things to supersede other things? Take a look at `arm/addsf3.S'.

johnwbyrd: Good catch — done in 6e48063. Dropped the `list(REMOVE_ITEM ... addsf3.c divsf3.c mulsf3.c subsf3.c)` block; since we have `mos/addsf3.S`, `mos/subsf3.S`, `mos/mulsf3.S`, `mos/divsf3.S`, `filter_builtin_sources` handles the removal automatically via base-name match (no `SUPERSEDES` needed here since the arch-specific and generic filenames align).

Verified end-to-end: configure now prints
```
-- For mos builtins preferring mos/addsf3.S to addsf3.c
-- For mos builtins preferring mos/divsf3.S to divsf3.c
-- For mos builtins preferring mos/mulsf3.S to mulsf3.c
-- For mos builtins preferring mos/subsf3.S to subsf3.c
```
the resulting `libclang_rt.builtins.a` contains `__addsf3/__subsf3/__mulsf3/__divsf3` only from `*.S.obj` members (no `.c.obj` versions built), and all four `Unit/*_test.c` binaries still pass under mos-sim.

The other `REMOVE_ITEM` block (SDK-provided `ashldi3.c` etc.) stays as-is — those have no matching `mos/*.S`, so `filter_builtin_sources` correctly leaves them alone and the explicit removal is still needed.

johnwbyrd:
Force-pushed `ed22351f` — a round of inner-loop tuning folded into the same commit.  All four unit tests still pass, and the changes were validated bit-for-bit against the pre-tuning code (result, NaN payloads, signed zeros, callee-saved zero page) over 62k operand pairs per routine on a cycle-accurate 6502 model.

Unit-test cycle counts under `mos-sim` (43/40/350+/351 vectors):

| routine  | before    | after     | delta |
|----------|-----------|-----------|-------|
| __addsf3 | 39,770    | 39,811    | +0.1% (branchless sticky-merge tradeoff; +~1cy/call on B_STK==0 path) |
| __subsf3 | 277,978   | 277,978   | 0 (trampoline into __addsf3) |
| __mulsf3 | 967,608   | 862,668   | -10.8% |
| __divsf3 | 535,174   | 476,967   | -10.9% |

Per-call on random operand pairs:

- `__divsf3`: 2231 -> 1890 cycles (-15%); `__mulsf3`: 1680 -> 1610 (-4%) on random, 1395 -> 754 (-46%) when one operand is an ordinary constant with trailing zero mantissa bytes (integer / simple fraction).

Notable micro-optimizations (all folded into the squashed commit; see updated commit body for the full list):

- Mul/div inner loops count down in `Y` (2-cy `DEY`) instead of `DEC` on a zp slot (5cy); `Y` is dead through both loop bodies.
- `__divsf3` decides whether to subtract with a 24-bit `CMP` (high byte first, early-out) and subtracts in place only when it commits, instead of subtracting into scratch and copying back.  Unifies the forced and elective subtract paths.
- `__mulsf3` skips a whole byte of iterations when the multiplier's low byte is zero -- shift multiplier down 8 bits, drop count 24->16, retained bits shift 8 fewer places, exactly compensating.  Two-level (skips 16 iterations when the low two bytes are zero).
- `__addsf3` parks the align residual count in `Y` (TAY/TYA = 4cy) instead of on the hardware stack (PHA/PLA = 7cy).
- `__addsf3` merges b's sticky into a's with branchless `ORA` -- legit because sticky is only ever tested for zero/nonzero downstream of the add path, never shifted bit by bit.

johnwbyrd:
Follow-up force-push `edc9efbe`: reverted the branchless-sticky-merge hunk in `__addsf3` and dropped the corresponding bullet from the commit body.  On reflection that change was a small cycle regression (+3cy on the common `B_STK == 0` path in exchange for -1cy on the truncation path) traded for aesthetic branchlessness — not worth it when the argument was only "cleaner," not "faster."

The other two `__addsf3` micro-optimizations (PHA/PLA -> TAY/TYA in the align byte loop, drop of the unused `TMP2` slot) are pure wins and remain.

Updated `__addsf3` unit-test cycle count: 39,811 -> **39,778** (essentially at the pre-tuning baseline of 39,770, +8cy / +0.2cy per call — noise).  Mul/div numbers unchanged.

mysterymath (approved):
LGTM!

PR #598, opened 2026-08-22
[MOS] Fix interrupt zero-page stack offset

### LINKED ISSUE #566: Broken register mapping into disallowed space (writing into $C5 on a C64)
I am compiling my code with (23.0.1)
mos-c64-clang++   -g -Os -DNDEBUG -Wall -pedantic -mcpu=mos6502x -std=gnu++20 
and link it with
mos-c64-clang++ -g -Os -DNDEBUG 
When I run it with --lto-emit-asm, for a specific part of the code, I get the following assembly
~~~.asm
; %bb.50:
        .loc    6 0 34                       
        ldx     #0
.LBB7_51:
        lsr     __rc2
        lda     __rc3
        ror
        cpx     #0
        bmi     .LBB7_53
; %bb.52:
        ldx     #0
        stx     __rc5
.LBB7_53:
        lsr     __rc2
        ror
        sta     mos8(.LEventDriverImp_zp_stk)   ; 1-byte Folded Spill
        .loc    6 759 38 is_stmt 1              
~~~

In my final .elf file, llvm-objdump shows me the following

~~~.asm
    16c5: a2 00         ldx     #$0
    16c7: 46 04         lsr     $4                      ; 0x4 <__rc2>
    16c9: a5 05         lda     $5                      ; 0x5 <__rc3>
    16cb: 6a            ror
    16cc: e0 00         cpx     #$0
    16ce: 30 04         bmi     $16d4 <function+0x253>
    16d0: a2 00         ldx     #$0
    16d2: 86 07         stx     $7                      ; 0x7 <__rc5>
    16d4: 46 04         lsr     $4                      ; 0x4 <__rc2>
    16d6: 6a            ror
    16d7: 85 c5         sta     $c5                     ; 0xc5 <__basic_zp_end+0x35>
  ~~~~

Accordingly, mos8(.LEventDriverImp_zp_stk)  seems to map to $c5, which is in my opinion not meant to be. The region between $90 and $FF is not included in the linker scripts and should be left completely untouched.

Btw, the function which contains the code is marked with __attribute__((interrupt_norecurse,nonreentrant)).

Once I remove nonreentrant, the access to $c5 disappears. If I ever add the nonreentrant attribute to another function, randomly locations betwenn $90 and $FF are used, eventhough (to my understanding) they are not mentioned for use in the linker file. 

Another observation: If I link with -mlto-zp=N and N>110, the linker complains that .zp overflows by some bytes. With 110, I get the described bahaviour (errornous access to $C5 with the nonreentrant attribute). With 109 (and lower), this behaviour disappears and nothing accesses $90-$FF anymore even when I label my function nonreentrant. If I label other functions with it though, access to the region $90-$FF is back.

I am not sure whether this is by chance, but if I look at the map-file without ((nonreentrant)) I get
~~~
      58       58       37     1 .zp
~~~
evaluating to the last byte being at $8F and with ((nonreentrant)) I get
~~~
      58       58       38     1 .zp
~~~
evaluating the last byte being at $90 (while my understanding is that $90 should already be excluded from __basic_zp.

Note that 0x58 + 109 is equal to $C5. (Which is strange because the wiki says that -mlto-zp sets the size of the available zero page excluding the imaginary registers. This would mean S=34+109=$8f => as expected?)

For me it seems there is a miscaclulation of the available zero-page space when ((nonreentrant)) is used.

Unfortunately, I do not understand enough of the internals of a compiler or linker to come up with something more useful at the moment, but I am willing to try help in any way I can to pin down what is happening.

Krows7:
`MOSZeroPageAlloc` uses `ZPOffset` as an offset into `zp_stack`, but interrupt offsets currently start at `GlobalZPSize + RegularZPSize`.

`zp_stack` only contains regular and interrupt stack storage. Promoted globals are allocated before it, so including `GlobalZPSize` in an offset relative to `zp_stack` counts that space a second time.

The existing `zp-alloc.ll` test already shows the same problem in a small case. It emits an 11-byte `zp_stack`, but places the one-byte interrupt stack at offset 12. Starting interrupt offsets at `RegularZPSize` moves it to offset 10.

I also verified the fix against the reproducer attached to #566. Its final zero-page layout is:

```text
.zp.data  $22, size 48
.zp.bss   $52, size 6
.zp       $58, size 56
```
The 54 bytes occupied before `.zp` give `GlobalZPSize = 54` for this allocation. In the patched LTO output, the regular stack uses 55 bytes and the one-byte `EventDriverImp` interrupt stack is placed at:

```asm
.L_Z14EventDriverImpv_zp_stk = .Lzp_stack+55
        .size   .L_Z14EventDriverImpv_zp_stk, 1
```

With the old calculation, the interrupt offset is `54 + 55 = 109`, giving `$58 + 109 = $c5`, exactly the invalid address reported in the issue. With the fix, the offset is 55, giving `$58 + 55 = $8f`, the last byte of the allocated `.zp` range.

The `zp-alloc.ll` expectation is updated accordingly. The updated test fails with the old implementation and passes with the fix. The full MOS CodeGen suite passes with 80 tests passed and 1 unsupported.

Fixes #566


mysterymath (approved):
Wow, niiiiice catch. Thanks!

PR #601, opened 2026-08-30
feat(MOS): DWARF register numbering and CFI support for source-level debugging

johnwbyrd:
## Summary

Enable LLDB source-level debugging on MOS 6502 by emitting DWARF CFI for the target's dual-stack unwinding model. Scoped to only what's needed for a debugger to produce a call stack and inspect locals — orthogonal work (builtins, stack-size diagnostics, obsolete AsmPrinter workaround cleanup) will land as separate PRs.

Alternative to #568 — uses MOS-only escape-based helpers instead of adding first-class expression variants to upstream `MCCFIInstruction`.

## What's in

Two commits:

1. **`feat(MC): MIR printer/parser support for DW_CFA_val_offset (NFC)`** — small upstream hygiene fix. `createValOffset` has been in the `MCCFIInstruction` C++ API but had no MIR printer/parser support, so serialization fell through to `<unserializable cfi directive>`. Adds the missing case in `MachineOperand.cpp` and matching lexer/parser rules alongside `OpOffset`/`OpRelOffset`. ~15 LOC across 4 files. No behavioral change to DWARF output; only makes MIR text serialization lossless.

2. **`feat(MOS): DWARF register numbering and CFI support`** — the debugging enablement:
   - DWARF register numbering: A(0), X(1), Y(2), P(3), S(4), PC(5), RC at 0x20000, RS at 0x30000
   - DwarfCFI as default `ExceptionHandling`, `CalleeSaveStackSlotSize = 1` (DWARF data alignment factor)
   - CIE initial frame state (CFA = normalized(S+3), PC = [normalized(S+1)])
   - Prologue CFI: switch CFA from hardware stack to soft stack, PC via `DW_CFA_expression`, S register `DW_CFA_val_expression`, callee-saved register rules
   - Epilogue CFI restore
   - `getFrameIndexReference()` so debuggers can locate local variables
   - Escape-based helpers in `MOSCFIExprHelpers.h` for the expression forms (compromise per llvm-mos/llvm-mos#568 discussion; if upstream ever adds first-class constructors, migration is a body swap on three helper functions with no call-site churn)

## Test plan

- [ ] `ninja check-llvm-codegen-mos` — CFI tests pass (`prologepilog.mir`, `prologepilog-ir.mir`, `zp-alloc.ll`, `frame-pointer-cfi.ll`)
- [ ] `llc -mtriple=mos -filetype=asm` on a small alloca+call function emits `.cfi_def_cfa`, `.cfi_val_offset`, `.cfi_escape 0x10, ...` (DW_CFA_expression), `.cfi_escape 0x16, ...` (DW_CFA_val_expression), `.cfi_endproc`
- [ ] MIR test check lines read as `val_offset $rs0, 0` (not `<unserializable cfi directive>`)

### DIFF
--- a/llvm/lib/Target/MOS/MCTargetDesc/MOSMCTargetDesc.cpp
+++ b/llvm/lib/Target/MOS/MCTargetDesc/MOSMCTargetDesc.cpp
@@ -17,10 +17,13 @@
 #include "MOSMCInstrAnalysis.h"
 #include "MOSTargetStreamer.h"
 #include "TargetInfo/MOSTargetInfo.h"
+#include "../MOSCFIExprHelpers.h"
 
-#include "llvm/BinaryFormat/ELF.h"
+#include "llvm/ADT/SmallString.h"
+#include "llvm/BinaryFormat/Dwarf.h"
 #include "llvm/MC/MCAsmBackend.h"
 #include "llvm/MC/MCCodeEmitter.h"
+#include "llvm/MC/MCDwarf.h"
 #include "llvm/MC/MCELFStreamer.h"
 #include "llvm/MC/MCInstrAnalysis.h"
 #include "llvm/MC/MCInstrInfo.h"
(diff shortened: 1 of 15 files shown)


llvm/lib/Target/MOS/MOSRegisterInfo.td
```diff
@@ -48,44 +55,58 @@ def sublsb : MOSSubRegIndex<1>;
...
-def N : MOSReg<8, "N">;
-def V : MOSReg<9, "V">;
-def Z : MOSReg<10, "Z">;
+// MOS 8-bit X register
+defm X : MOSReg8<1, "x">;
+
+// MOS 8-bit Y register
+defm Y : MOSReg8<2, "y">;
+
+// MOS 8-bit stack register.  Some debuggers falsely claim this to be
```
mysterymath: "Some debuggers"? Huh?

llvm/lib/Target/MOS/MCTargetDesc/MOSMCTargetDesc.cpp
```diff
@@ -17,10 +17,13 @@
 #include "MOSMCInstrAnalysis.h"
 #include "MOSTargetStreamer.h"
 #include "TargetInfo/MOSTargetInfo.h"
+#include "../MOSCFIExprHelpers.h"
```
mysterymath: nit: `../` is unnecessary

PR #602, opened 2026-09-02
feat(lldb): add MOS ABI plugin

johnwbyrd:
## Summary

Add the LLDB ABI plugin for MOS 6502. Third piece of the debugging trio (with #571 and #601) that lets LLDB understand MOS programs.

- `ABISysV_mos` — main plugin. Looks up DWARF register numbers dynamically via `MCRegisterInfo::getDwarfRegNumByName`, so the plugin automatically tracks whatever `MOSRegisterInfo.td` says. Rebuilds the same normalized hardware-stack expression `((S+N) & 0xFF) | 0x0100` (CFA=S+3, PC at S+1, S_prev at S+2) that the CIE in #601 emits — compiler and debugger are two ends of the same design.
- `MOSImaginaryRegisters` — RC0-RC255 byte pseudo-registers and RS0-RS127 pair pseudo-registers. Not reported by the GDB stub; the plugin reads them from target memory using addresses scanned out of ELF `__rcN` symbols.
- `MOSRegisterContext` — presents the combined view (physical from stub + imaginary from memory) to LLDB.
- Entirely under `lldb/source/Plugins/ABI/MOS/` except one line in the parent `CMakeLists.txt`.

## Dependencies

- **Requires #571** (feat(lldb): ABI extension points for custom register contexts). Uses `CreateRegisterContextForThread` and `ProvidesRegisterInfoOverride` from that PR. This PR is opened against a branch stacked on #571 and should land after #571 merges.
- **Pairs with #601** (feat(MOS): DWARF register numbering and CFI). Code-wise independent, but the two together are what enables end-to-end debugging. This plugin consumes the DWARF #601 emits.

## Test plan

- [ ] `ninja lldb` builds cleanly with #571 as base (verified locally against rebased #571 + current upstream/main)
- [ ] `lldb foo.o` (a MOS `.o` compiled with `-g` via clang/#601's llc) recognizes the architecture as `mos` and finds `main`
- [ ] `image lookup -n main --verbose` shows the correct DWARF register numbers for arguments and locals
- [ ] Attaching to MAME's GDB stub while running a compiled MOS program produces correct backtrace, source-level stepping, and local variable inspection

### DIFF
--- a/lldb/include/lldb/Target/ABI.h
+++ b/lldb/include/lldb/Target/ABI.h
@@ -179,6 +179,22 @@ public:
     return std::nullopt;
   }
 
+  virtual lldb::RegisterContextSP
+  CreateRegisterContextForThread(
+      lldb_private::Thread &thread,
+      uint32_t concrete_frame_idx) const {
+    // Default: return nullptr, meaning use the default RegisterContext
+    return nullptr;
+  }
+
+  // Returns true if the ABI is prepared to provide authoritative register info for override.
+  virtual bool ProvidesRegisterInfoOverride() const { return false; }
+
+  // If opted in, returns a fully-populated Register for the given name, or std::nullopt if not recognized.
+  virtual std::optional<lldb_private::DynamicRegisterInfo::Register>
+  GetCanonicalRegisterInfo(llvm::StringRef name) const { return std::nullopt; }
+
+
 protected:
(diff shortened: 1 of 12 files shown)


johnwbyrd:
Folded into #571 per @mysterymath's review feedback that the API additions and their caller should be reviewed together. The plugin commit is now the second commit on the #571 branch.

